Issue #3, opened 2016-08-14
Automatically create an apk after PRs with travis


mariobehling:
Please add an automatically generated apk with travis into the repository.

pa1pal:
#10

PR #10, opened 2016-08-17
Upload sample apk on every commit.

pa1pal:
Added **upload-apk.sh** which copies the generated apk file in sample apk directory.
Thanks @the-dagger for script
#3


mariobehling:
Please help to add this to the travis.yml in following steps. So the build is tested. Cheers.

PR #14, opened 2016-08-22
Camera bugs fixed

pa1pal:
- Changing front/ back camera now working in every phone.
- app not crashing after image click.
- Filters working.
#15


mariobehling:
Good job! Keep it up!

PR #21, opened 2016-09-13
Wordpress upload working, tested on Phimpme blog

pa1pal:
Tested wordpress upload on phimpme wordpress test blog [https://github.com/fossasia/phimpme-wordpress](url)

According to me we should keep the implemented for wordpress as it is.It is good and will work for us.

**Screenshots of the app and wordpress.**
phimpmewordpress
wordpress


mariobehling:
Great! Thank you.

PR #43, opened 2017-03-01
fix: App crashing on adding picture at Upload

CloudyPadmal:
The query handler used which is **managedQuery()** is deprecated! Replaced it with _**getContentResolver().query()**_

### DIFF
--- a/app/src/main/java/vn/mbm/phimp/me/PhotoSelect.java
+++ b/app/src/main/java/vn/mbm/phimp/me/PhotoSelect.java
@@ -49,7 +49,7 @@ public class PhotoSelect extends Activity {
 		showFromCache();		
 		final String[] data = { MediaStore.Images.Media.DATA };
 		final String orderBy = MediaStore.Images.Media._ID;
-		pathcursor = managedQuery(
+		pathcursor = getContentResolver().query(
 				MediaStore.Images.Media.EXTERNAL_CONTENT_URI, data,
 				null, null, orderBy);	
 		final Button selectBtn = (Button) findViewById(R.id.selectBtn);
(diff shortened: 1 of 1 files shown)


app/src/main/java/vn/mbm/phimp/me/PhotoSelect.java
```diff
@@ -309,7 +309,7 @@ public void run() {
 	 		        
 	 			}
 	    	}catch(RuntimeException runex){
-	    		this.onCancelled();
+	    		//this.onCancelled();
```
Abhi2424shek: Should this be uncommented, I guess it is a good UI to stop the process when an exception has occurred.

CloudyPadmal:
@Abhi2424shek I've uncommented it and made the PR again! I referred to this [link](http://stackoverflow.com/questions/24463364/android-activity-asynctask-and-oncancelled) when I commented it! But I agree to what you say! :D Can you review this?

CloudyPadmal:
@Abhi2424shek There were few other lines where the deprecated code needed to be replaced in the same class file! I've changed them and squashed them together!

app/src/main/java/vn/mbm/phimp/me/PhotoSelect.java
```diff
@@ -281,7 +281,7 @@ protected String doInBackground(Cursor... cursor) {
 	 				boolean check = PhimpMe.cache.check(path);
 	 				if(!check){	 					
 	 					@SuppressWarnings("deprecation")
```
Abhi2424shek: Since you removed the depreciated code can you remove the supress tag as well as we don't need to supress the depreciation as there w9uld be no depreciation!

CloudyPadmal: Sure @Abhi2424shek !! PR is right away :D

CloudyPadmal:
@Abhi2424shek Thank you!! :1st_place_medal:

PR #63, opened 2017-03-03
Fixes issue #54 Add PR and Issue templates

### LINKED ISSUE #54: Add PR and Issue templates
Adding PRs and issue templates would make the project more better, we could make use of the default .github folder to do so. Can we have it?

anantprsd5:
Fixes #54 Added PR and Issue templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,0 +1,23 @@
+**Actual Behaviour**
+
+Please state here what is currently happening.
+
+**Expected Behaviour**
+
+State here what the feature should enable the user to do.
+
+**Steps to reproduce it**
+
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,23 @@
+**Actual Behaviour**
+
+Please state here what is currently happening.
```
Abhi2424shek: Would it be better to these a comment?

anantprsd5: These comments gives a clear idea of what to write inside these templates. So IMO its better to have it there.

Abhi2424shek: Ok then!

Issue #69, opened 2017-03-03
App crashes while trying to switch camera


anantprsd5:
**Actual Behaviour**

Currently when the user tries to switch the camera, there is always an app crash as Camera API is depreciated

**Expected Behaviour**

There should be no such crashes and depreciated code should be changed with new cameraV2 api.

**Steps to reproduce it**

Open app , go to camera , press the camera switch button.

**Would you like to work on the issue?**

Yes.

dynamitechetan:
on my phone its working perfectly 
Any logs?

anantprsd5:
```
E/AndroidRuntime: FATAL EXCEPTION: main
                                                                 Process: vn.mbm.phimp.me, PID: 12855
                                                                 java.lang.RuntimeException: setParameters failed
                                                                     at android.hardware.Camera.native_setParameters(Native Method)
                                                                     at android.hardware.Camera.setParameters(Camera.java:2173)
                                                                     at vn.mbm.phimp.me.Preview.switchCamera(Camera2.java:597)
                                                                     at vn.mbm.phimp.me.Camera2$4.onClick(Camera2.java:221)
                                                                     at android.view.View.performClick(View.java:5226)
                                                                     at android.view.View$PerformClick.run(View.java:21350)
                                                                     at android.os.Handler.handleCallback(Handler.java:739)
                                                                     at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                     at android.os.Looper.loop(Looper.java:148)
                                                                     at android.app.ActivityThread.main(ActivityThread.java:5582)
                                                                     at java.lang.reflect.Method.invoke(Native Method)
                                                                     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616
```

anantprsd5:
@Abhi2424shek @pa1pal Can you please check if the app is crashing on camera switch in your device because for me its crashing but for @dynamitechetan its working perfectly. If the crash is there, I will start working on it.

pa1pal:
@anantprsd5 it's not crashing in mine. But if this in your case. So you can fix it.

Abhi2424shek:
I am not too getting a crash! Can you check your apk version. @anantprsd5

anantprsd5:
[...]

PR #76, opened 2017-03-04
Fix: Crash on switching camera

dynamitechetan:
Fixes issue #69


Abhi2424shek:
@dynamitechetan always refeR the issue number in the commit.

PR #96, opened 2017-03-08
Added slack id as an environment variable. Closes #95

### LINKED ISSUE #95: Disable pushes to slack for forks
**Actual Behaviour**

When a commit is made on a fork, slack gets notification.

**Expected Behaviour**

Notification should not be made.

Yes.

Abhi2424shek:
Closes #95.
The environment variable has been added to the dashboard.
Now notif won't go to slack from forks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,4 +44,4 @@ after_success:
   - 'if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./upload-apk.sh; fi'
 
 notifications:
-  slack: fossasia:JgzycrBUs0nKnmJhsAxCB4bL
+  slack: $SLACK_ID


unitware:
Did this work for you? I tried it in another project earlier today without success.

here: https://github.com/unitware/cleanflight/blob/beta-test-env-var-in-travis-yml-to-silence-private-notifications/.travis.yml

Abhi2424shek:
@unitware it works when you are on the correct branch, I think it should. Have you added the variable? It doesnot work on forks!

unitware:
I have added the variable in my travis account, as an environment variable. 

image

And also in the travis.yml file. 
image

Did I miss anything?

Abhi2424shek:
I don't think so! @unitware try adding something like betaflightgroup:$SLACK_ID
Where slack id is the rest after the :

unitware:
But that is not the way you did in this PR?

I'll try it anyway

Abhi2424shek:
This should work, see: https://docs.travis-ci.com/user/notifications/#Configuring-slack-notifications

I am on mobile, since not sure I would just close the PR and make another one when I am on PC.
Apologies sorry.

Issue #86, opened 2017-03-07
User Interaction Issues 


AliAyub007:
**Actual Behavior**

If you press back button on settings, upload and camera it goes back to gallery and then close the app promptly and we have to select the options from navigation given below. There are some Spelling and Grammar error in dialog boxes.

**Expected Behaviour**

It should go back to home screen and shouldn't close app. It can be made swipe view that if screen is swiped left then it goes to camera option and if swiped again it goes to upload and so on 

**Steps to reproduce it**

It might because of value which remain saved in key code it can be made right by making it to "0" after setting view to home screen. 
Screen can be converted into swipe view by viewpager. 

**Would you like to work on the issue?**

Yes I would like to work on these.

dynamitechetan:
Read about Bottom Navigation Bar guidelines. 
Using swipe gestures on the content area does not navigate between views.
So no need to implement swipe gestures. You can work on the grammar and other issues.

Abhi2424shek:
Yes pls work only on the grammar and english issues!

Sanjayjeyavel:
Hey! I'm new to open source and would love to work on this!!

AliAyub007:
Okay I will do this till tonight.

AliAyub007:
@sanjayjeyavel If you are new to GitHub try taking tutorials they will you alot. You have to learn git for this and here is a link I think this will be useful kindly check this. 
https://idope.se/torrent/apps/https:/www.idope.se/torrent/essential/1c0c89e39140f598d30e71f42806d638e83626ef/

Sanjayjeyavel:
@AliAyub007 I will!! Thank you.

pa1pal:
@AliAyub007 The link you shared in your comment is torrent of a Paid online course. The torrent of copyrighted and paid material should not be promoted, it is completely illegal.
And that too on platform like GitHub. I updated your comment, don't do this from next time.

AliAyub007:
@pa1pal  I am really sorry I didn't know this and I won't do this next time.

Abhi2424shek:
Dont make multiple PRs for a small issue!

PR #97, opened 2017-03-08
Update Styles.xml

AliAyub007:
Removing Title bar and giving app a professional and better user interactive interface.

Fixes issue #86 
Changes: Removing Title bar and giving app a professional and better user interactive interface.

Screenshots for the change:

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -2,9 +2,12 @@
 
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Phimpme.Base">
+        <item name="colorPrimary">@color/colorPrimary</item>
+        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
+        <item name="colorAccent">@color/colorAccent</item>
     </style>
 
-    <style name="Phimpme.Base" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="Phimpme.Base" parent="Theme.AppCompat.Light.NoActionBar">
         <item name="colorPrimary">@color/colorPrimary</item>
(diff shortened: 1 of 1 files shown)


dynamitechetan:
Action bar is an important design element that provides a consistent familiar look between Android apps. It also provides a space for the app or activity’s identity, thus enabling the user to know their location in the app, and easy access to the actions that can be performed.
I think it should be there in the app.

AliAyub007:
I was thinking it in a way that there is a navigation bar in app which is showing position to user so that's why I thought of it.

dynamitechetan:
But there are other activities like Crop, Edit Image etc. which will require action bar. So why not to have a consistent look for the entire app.

AliAyub007:
yeah that is a valid point. (Y)

app/src/main/res/values/styles.xml
```diff
@@ -2,9 +2,12 @@
 
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Phimpme.Base">
+        <item name="colorPrimary">@color/colorPrimary</item>
+        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
+        <item name="colorAccent">@color/colorAccent</item>
     </style>
 
-    <style name="Phimpme.Base" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="Phimpme.Base" parent="Theme.AppCompat.Light.NoActionBar">
```
Abhi2424shek: We want an action bar for sure.

Abhi2424shek:
I am closing this PR as the change is not needed! I hope you understand, next time comment on a relevant issue before making a PR.

PR #98, opened 2017-03-08
Update PhimpMeGallery.java

AliAyub007:
Correction of grammatical error in dialog ox after uploading at line 116.

Fixes issue #86 

Changes: Fixed Grammatical error in Dialog box after uploading photo. 

Screenshots for the change:

PR #100, opened 2017-03-08
Update Phimpme.java

AliAyub007:
Setting keycode to zero after getting view of home screen (Gallery) on pressing back button so that it shouldn't close the app after getting back to home screen at pressing back once on any other option. It should only return to home screen.

Fixes issue #86 

Changes: Fixed error that if you press back button on any option it goes to home screen and closes app promptly now it just goes to home screen. 

Screenshots for the change:


pa1pal:
@AliAyub007 Do this change in your #98. So this will reflect complete changes for issue #86

AliAyub007:
Done with that.

Issue #116, opened 2017-03-11
Share App feature


rohanagarwal94:
**Actual Behaviour**

No sharing of app right now.

**Expected Behaviour**

User will be able to share the app on various other apps. Button will be located in settings.

**Would you like to work on the issue?**

Yes, I would like to work on it.

Abhi2424shek:
Ok pls go on.

PR #136, opened 2017-03-14
#116 issue solved:Share app Feature added

optnio00:
Fixes issue #116

Changes: [Add here what changes were made in this issue and if possible provide links.]

Screenshots for the change: 
screenshot_2017-03-14-17-44-14
screenshot_2017-03-14-17-44-22


dynamitechetan:
I think it's not the correct placement for sharing option. We can add no more than 5 options in bottom navigation bar and we have map PR pending which is a important feature of the app. So place this somewhere else but not the bottom bar.

optnio00:
ok i change the position . Now i will put it in settings tab

Abhi2424shek:
Dont put it on settings.

rohanagarwal94:
@optnio00 it is quite derogatory if you solve issues assigned to/ opened by other fellow contributors without their permission. @Abhi2424shek if you only could keep a check on this.

heysadboy:
@optnio00 I think you should put share option in toolbar. In the for of those three dots on the right end.

PR #138, opened 2017-03-14
Making use of toolbar

heysadboy:
1. Adding the refresh button on toolbar so the grid will have more space. 

2. Removing the extra bar containing title and button. 

3. No toolbar in camera and map feature to make the view larger.

screenshot_2017-03-15-00-22-09


pa1pal:
@heysadboy can you add a click effect on the button.

heysadboy:
@pa1pal Sure.

PR #169, opened 2017-03-18
Fixes #163: Improved Layout

### LINKED ISSUE #163: Improvement in Preview Camera Layout
**Actual Behaviour**

The flash button and switch camera icon takes up a lot more space in preview fragment

**Expected Behaviour**

Their size should be reduced and placed at corner for better preview 

@Abhi2424shek Yes i would like to work on this issue.

madhavrathi:
Fixes issue #163 

Changes:Reduced the size of photobuttons and cornered them

Screenshots for the change: 
screenshot_2017-03-18-17-21-40-028_vn mbm phimp me


CloudyPadmal:
How about making them a little bit bigger and align them both side of the screen? Cam to the left and flash to the right? :innocent:

madhavrathi:
@CloudyPadmal Hey, i believe that the size is perfect as we want the end user to have maximum area for preview of image and at the same time not ignore these buttons. About the alignment issue i think it's better to have them on single side as the icon for camera shift is having large transparent portion so it'll mix up with bright images.
@Abhi2424shek what do you think ?

app/src/main/res/layout/camera.xml
```diff
@@ -14,23 +14,21 @@
...
-            android:layout_marginTop="16dp"
-            android:gravity="center" >
+            android:layout_marginTop="16dp">
 
             <ImageButton
                 android:id="@+id/switch_camera"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:layout_gravity="center_horizontal|center"
+                android:layout_width="20dp"
```
pa1pal: Just giving small hardcoded value will not solve the problem.
Best is if you set it programmatically according to the screen size.
Also always add any dimension in `dimen.xml`

Abhi2424shek:
Make the first icon a bit of logical and address the issue highlighted by pa1pal.

app/src/main/res/layout/camera.xml
madhavrathi: @pa1pal @Abhi2424shek  i have now added the dimensions in dimen.xml and about setting them programtically, I think someone is working on layout for Camera UI, so we'll add the code in that PR.

madhavrathi:
screenshot_2017-03-19-15-23-44-876_vn mbm phimp me
@Abhi2424shek  hey check out the new icon for camera switch. I guess it is pretty clear now.
Should i send updated PR ?

Abhi2424shek:
Make it of the same size as the flash button.

madhavrathi:
screenshot_2017-03-19-18-47-59-340_vn mbm phimp me 1
@Abhi2424shek  There were some issues with previous icon's resolution. Check this one.

app/src/main/res/values/dimens.xml
```diff
@@ -5,6 +5,7 @@
   <dimen name="fab_margin">16dp</dimen>
   <dimen name="headingText">16sp</dimen>
 
+  <dimen name="cameraIcon">30dp</dimen>
```
pa1pal: Still the hardcoded dimensions. This will not solve the problem. Dimensions should be done pro-grammatically. Just a few more lines needed.

madhavrathi: @pa1pal Ok. i will set them pro-grammatically and send a PR for this. Thanks for Review

Issue #150, opened 2017-03-16
Max file size issue fix


rohanagarwal94:
**Actual Behaviour**

Maximum file size cannot be changed in settings.

**Expected Behaviour**

The user should be able to change the maximum file size by clicking the image button

**Screenshots of the issue**

![screenshot_2017-03-16-22-27-11](https://cloud.githubusercontent.com/assets/13241256/24008626/258ad526-0a98-11e7-81bf-86818456488a.png)

**Would you like to work on the issue?**

Yes, I am already working on it.

Abhi2424shek:
Are you mentioning of the photo save size?

rohanagarwal94:
oh yes. The settings in the above image to change max file size

Abhi2424shek:
Ok then. Pls do it.

PR #179, opened 2017-03-19
Max file size issue fix

rohanagarwal94:
Fixes issue #150 

Screenshots for the change: 

screenshot_2017-03-19-15-36-45


Abhi2424shek:
Can we have a text box along with them where the user can enter his choice?

rohanagarwal94:
@Abhi2424shek I guess that will make a simple task complicated. I dont think user needs to put in specific size. @pa1pal thoughts?

pa1pal:
I agree a quick selection option would be better. But using spinner here would be good choice instead of radio buttons in dialog.

rohanagarwal94:
@pa1pal That will interfere with the basic UI of the settings fragment since majorly all the preferences will have dialog. Moreover it would hide the delete the photos preference which would look odd. Suggest otherwise if you still think that would work. Will update the PR then.

screenshot_2017-03-16-22-27-11

Abhi2424shek:
We would go with this then.

app/src/main/java/vn/mbm/phimp/me/Settings.java
```diff
@@ -155,8 +159,10 @@
 	File tmp_folder;
 	int error_count = 0;
 	ProgressDialog pro_gress;
+    private AlertDialog maxSizeDialog = null;
```
pa1pal: please indent whole code.

rohanagarwal94:
@pa1pal made the necessary changes. Please review.

Issue #180, opened 2017-03-19
No image selected label is not shown in Upload Photos list


CloudyPadmal:
**Actual Behaviour**

When images are added to the upload photos list and in case if we remove all of them, _No Picture Selected :(_ label is invisible.

**Expected Behaviour**

When the list is emptied, _No Picture Selected :(_ label should appear again!

**Steps to reproduce it**

1. Open to app
2. Go to _Upload_ tab
3. Add some images to upload photos list
4. Long click one image and then select all images
5. Remove selected images

**LogCat for the issue**

> Not available

**Screenshots of the issue**

![device-2017-03-19-164625](https://cloud.githubusercontent.com/assets/14261304/24080359/a00e2cee-0cc3-11e7-9e07-d2591d2ad1b2.png)

**Would you like to work on the issue?**

Yes :)

PR #181, opened 2017-03-19
fix: [ #180 ] No photos label will appear again!

CloudyPadmal:
Fixes issue #180 _No Picture Selected :(_ label invisible after removing all the images

Changes: 
- Made _No Picture Selected :(_ label visible when the list is emptied
- Wrote Espresso tests to validate

Screenshots for the change: 
device-2017-03-19-1


CloudyPadmal:
@Abhi2424shek Thank you! ^_^

PR #193, opened 2017-03-19
Updated Readme with new Screenshots

dynamitechetan:
Updated Readme with new Screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,19 +28,19 @@ We have the following branches
 
   <table>
     <tr>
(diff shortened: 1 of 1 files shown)


heysadboy:
We have upcoming UI changes. It'd be great if you could add these when it reaches a certain point.

dynamitechetan:
yes. But map option in the bottom bar is there for a while now. So I thought to update them.

PR #433, opened 2017-05-11
Removed unnecessary code from the repo. Setup a new boilerplate.

pa1pal:
Removed unnecessary code from the repo. Setup a new boilerplate to start on. Right now app has a bottomNavigationView with three tabs only. 
We will code now on above it.


pa1pal:
@mariobehling please approve this. It needs to force merge. I will do this on command line.

Issue #483, opened 2017-05-25
Shifting ImageEditingLibrary from module to package


pa1pal:
**Actual Behaviour**

Currently, `imageEditingLibrary` is added as a separate module. 

**Expected Behaviour**

Shift this into packages as we added `leafpic` and `opencamera` for using common utilities.

**Would you like to work on the issue?**

Yes

mohitmanuja:
@pa1pal Also do the toolbar and bottomnavigation color as appPrimaryColor.

PR #488, opened 2017-05-26
shifted editimagelibrary module code to main app

pa1pal:
Fixes issue #483 

Changes: Moved all files and added fields required by `editImageLibrary`

@mohitmanuja @anantprsd5 @Subhankar29 Please review.


app/src/main/java/vn/mbm/phimp/me/editor/editimage/EditImageActivity.java
```diff
@@ -189,7 +197,9 @@ public void onClick(View v) {
         mRotateFragment = RotateFragment.newInstance();
         mAddTextFragment = AddTextFragment.newInstance();
         mPaintFragment = PaintFragment.newInstance();
-
+        banner.setBackgroundColor(themeHelper.getPrimaryColor());
```
mohitmanuja: It should be write in onResume if in case user change the theme and this activity is in the pause state at that time.

pa1pal: ok, I will shift this in `onResume`. But I think that case is not possible with `EditImageActivity` for now, settings changes required activity to get destroyed.
But for future we can add it to onResume.

mariobehling:
If this is WIP, how can it be approved?

pa1pal:
@mariobehling it was working fine. But later ndk configuration bug found. I'm fixing it now.

Issue #496, opened 2017-05-29
Update README.md


anantprsd5:
**Actual Behaviour**

Old feature information and screenshots are used in the file.

**Expected Behaviour**

Feature table and the screenshots should be updated.

**Would you like to work on the issue?**

Done

PR #497, opened 2017-05-29
Updated Screenshots and feature table in README.md

anantprsd5:
Fixes issue #496 Update README.md

Changes: Added new screenshots and updated the feature table.


Subhankar29:
I will suggest removing About App screen shot because it is not a feature.

Issue #530, opened 2017-06-03
On tap on reset setting the activity should be closed


mohitmanuja:
**Expected Behaviour**

On tap on reset setting the activity should be closed. It can be checked by the back press after reset.

**Steps to reproduce it**

Open camera setting and click on reset settings then back press.

**Would you like to work on the issue?**

Yes.

Issue #532, opened 2017-06-04
Handling state of activity through bottom navigationbar


mohitmanuja:
**Actual Behaviour**

Activity is opened again when we are at 'home' and click on 'home' again.

**Expected Behaviour**

If the activity is in focus already then we should not open it again.

**Steps to reproduce it**

Click on camera and again click on the camera. Similarly with home and accounts.

**Would you like to work on the issue?**

Yes.

PR #533, opened 2017-06-04
Cleared all activity stack history on reset

mohitmanuja:
Fixes issue #530 
Changes:

On reset -
Cleared all activity stack history and re-opened the app.
It can be checked by the back press after reset.


mohitmanuja:
@pa1pal Yes, Right now I am clearing all the stack history of activity and re-opening this app or we can call finish on setting only that will also work.
So should I finish settingActivity instead of clearing all stack?

pa1pal:
@mohitmanuja Clearing stack is one option or some other way to prevent re opening the same activity again on top of it.

mohitmanuja:
@pa1pal On click on reset settings in camera settings we have to close that activity right?

pa1pal:
Yes

mohitmanuja:
So simply we can call finish(); beacause it is in settingActivity at that time.

pa1pal:
for that yes.

mohitmanuja:
So that is what I am saying instead of clearing stack we can do finish() , I will update my PR and will call finish() on reset to finish the settingActivity.

pa1pal:
@mohitmanuja yes for settings is fine. 
I confused in which context you asked. But add that part as well, which I commented. For not adding activity over the same activity.

mohitmanuja:
That is the different issue #532 to handle the activity stack @pa1pal. and yes I will do as you are saying If(item.IsChecked()).

mohitmanuja:
So I think on reset clearing all stack and the opening app is again fine? (Did in this PR)
or should I finish settingActivity only.

pa1pal:
opening app again is fine.

mohitmanuja:
Okay, that is done. Review this PR.

Issue #525, opened 2017-06-03
Toggle button to choose between different colour effects


Subhankar29:
**Actual Behaviour**
Currently, we have to go to options choose between different colour modes. For a user to change multiple times between the different colour modes, it will be a tedious process.

**Expected Behaviour**

Add a toggle Button on the camera screen to change between different colour modes. 

**Screenshots of the issue**

![screenshot_2017-06-03-17-45-04-273](https://cloud.githubusercontent.com/assets/17337006/26753489/15b8e798-4885-11e7-852c-377384929d2c.jpeg)

**Would you like to work on the issue?**

Yes

Subhankar29:
@pa1pal, @mohitmanuja, @vinaysajjanapu @anantprsd5 your suggestions?

vinaysajjanapu:
On the screen?

Subhankar29:
Yes. I was thinking on the top right to look clean.

Subhankar29:
or left.

anantprsd5:
@subhankar29 Yes, put that at top left. What to do with the options in popupView then?

Subhankar29:
Let's leave it as it is because of quicker selection for the user to choose a specific effect.

PR #536, opened 2017-06-05
Added button to switch between different color effects

Subhankar29:
Fixes issue #525 

Changes: Added button to switch between different color effects on the camera screen.


Subhankar29:
@anantprsd5 The popup View is now working.

Subhankar29:
@pa1pal I believe as a user you don't know which effect will look nice on the photo you are clicking. And a user won't know how the effects look by their name. It is also a tedious process to go through the popview each and every time a user wants to change the color effect. And like Snapchat it will be easier for the user to choose between the effects.

app/src/main/java/vn/mbm/phimp/me/opencamera/Preview/Preview.java
```diff
@@ -1372,7 +1372,27 @@ public void onFaceDetection(CameraController.Face[] faces) {
...
-			CameraController.SupportedValues supported_values = camera_controller.setColorEffect(value);
+            final int[] colorNumber = {0};
+            CameraActivity.toggle.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    String[] colorEffect = getResources().getStringArray(R.array.color_effects);
+                    colorNumber[0]++;
+                    if(colorNumber[0] == 15)
+                        colorNumber[0] = 0;
+                    CameraController.SupportedValues supported_values = camera_controller.setColorEffect(colorEffect[colorNumber[0]]);
```
mohitmanuja: You can get a list of supported color by using **preview.getSupportedColorEffects()**  instead of hardcoding the effects because may be some effect will not work on any other phones.

Subhankar29: Oh, thanks for this one. Was wondering for a while.

pa1pal:
@Subhankar29 okay. fine in that case.
Although I didn't click much to change. But I think focus is creating some hindrance in between.

PR #539, opened 2017-06-05
Managed the activity stack on same item click on bottom navigation bar.

mohitmanuja:
Fixes issue #532 

Changes: 

Managed the activity stack on clicking on the same item of the bottom navigation bar.
You are at 'home' and if you click on 'home' button again, activity will not open again.


app/src/main/java/vn/mbm/phimp/me/base/BaseActivity.java
```diff
@@ -43,14 +43,19 @@ public void onPause() {
     public boolean onNavigationItemSelected(@NonNull final MenuItem item) {
         switch (item.getItemId()) {
             case R.id.navigation_home:
-                startActivity(new Intent(this, LFMainActivity.class));
-                finish();
+                if (getNavigationMenuItemId() != R.id.navigation_home) {
```
pa1pal: Why checking each item by name, if you can get this by `item.getItemId()`. It will make code not redundant. 
add this check in the beginning and execute `switch` block on the basis of that.

mohitmanuja:
@pa1pal Thanks for the suggestion. Review it again.

Issue #537, opened 2017-06-05
Media count in the Album is always zero


anantprsd5:
**Actual Behaviour**

At the first screen of the app, the media count of an album is always displayed 0 in spite of it containing media items.

**Expected Behaviour**

It should display the actual amount of media in an Album

**Screenshots of the issue**
![screenshot_20170605-222617](https://cloud.githubusercontent.com/assets/20367966/26794233/8338118a-4a3e-11e7-8ba7-f17c1d87b51b.jpg)

**Would you like to work on the issue?**

Yes.

PR #540, opened 2017-06-05
Fixed media count issue

anantprsd5:
Fixes issue #537 Media count in album is always zero

Changes: Now the count is displayed according to no. of media in the album

### DIFF
--- a/app/src/main/java/vn/mbm/phimp/me/leafpic/data/providers/MediaStoreProvider.java
+++ b/app/src/main/java/vn/mbm/phimp/me/leafpic/data/providers/MediaStoreProvider.java
@@ -126,8 +126,7 @@ public class  MediaStoreProvider {
 
 	private static int getAlbumCount(Context context, long id) {
 		int c = 0;
-		String selection = "( "+MediaStore.Files.FileColumns.MEDIA_TYPE + "=? or "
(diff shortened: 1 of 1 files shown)


app/src/main/java/vn/mbm/phimp/me/leafpic/data/providers/MediaStoreProvider.java
```diff
@@ -130,6 +130,7 @@ private static int getAlbumCount(Context context, long id) {
 										   + MediaStore.Files.FileColumns.MEDIA_TYPE + "=? ) and " + MediaStore.Files.FileColumns.PARENT + "=?";
 
 		String[] selectionArgs = new String[] {
+						String.valueOf(MediaStore.Files.FileColumns.MEDIA_TYPE),
```
vinaysajjanapu: @anantprsd5 The issue got fixed but can you explain what exactly have you done here?

anantprsd5: @vinaysajjanapu I don't know how the query table is working but I observed that the 1st index of selectionArgs is accounting for the media count, so adding the 0th index as media_type makes the second index as media_type_image which fixes the issue.

anantprsd5:
@vinaysajjanapu I have found other way to solve it, please have a look at the changes.

Issue #544, opened 2017-06-06
Add option to share the application via Firebase App invite


anantprsd5:
**Actual Behaviour**

No application sharing option is there

**Expected Behaviour**

Add option to share the application via Firebase App invite

**Would you like to work on the issue?**

Yes

PR #545, opened 2017-06-06
Added App invite via Firebase deep link

anantprsd5:
Fixes issue #544 App option to share phimpme.

Changes: Added option to share phimpme via deep link which opens the application if it is installed in the phone or downloads the build via Github link which can later be replaced by Playstore.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,7 +35,7 @@ android {
     }
 
     configurations {
-        compile.exclude group: 'org.jetbrains' , module:'annotations'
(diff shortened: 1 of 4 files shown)


the-dagger:
@anantprsd5 I'll suggest that you share the APP using Firebase App invite instead.
That is a better way than sharing the direct apk instead.

anantprsd5:
@the-dagger Sure. Will add it using Firebase App invite then. Will make the changes soon.

Issue #598, opened 2017-06-15
Toggle button in the camera doesn't hide on taking picture


anantprsd5:
**Actual Behaviour**

Just like other buttons of the application, toggle button should hide itself on clicking the take picture button.

**Expected Behaviour**

1. Hide the button while taking the photo.
2. Remove the pop up menu options for photo effects.
3. Add showcase view to show toggle button functionality.
4. Add reset function on long pressing the toggle button.

**Steps to reproduce it**

1. Open camera.
2. Take photo 

**Would you like to work on the issue?**

Yes

mohitmanuja:
@anantprsd5 I think we should remove this button as well as effects from the popup. @pa1pal @vinaysajjanapu @Subhankar29 @anantprsd5. Because we have better effects in our edit activity.
Need suggestions.

vinaysajjanapu:
@anantprsd5 ok that buttons should be hidden during photo capture.

@mohitmanuja I don't think we have to remove that button. Most people (including me) tend to use live filters(if they are good). So we will keep that button as it is and change the live filter types later on. 
We can remove that list of effects in popup as it is redundant.

anantprsd5:
@vinaysajjanapu @mohitmanuja If we are keeping the toggle button then we also have to keep the pop up menu options so that user can cancel any effects easily or else they have to press the toggle button 14-15 times to return to normal mode.

Subhankar29:
@mohitmanuja Even I think we should keep them. We can always improvise the effects. Plus user will know the look of the photo before clicking it.

vinaysajjanapu:
@anantprsd5 we can reset filter when filter button in long pressed.

anantprsd5:
@vinaysajjanapu But how will the user know that effects will be reset on long press. I feel it is more user friendly their in the pop up settings. Suggestions?

vinaysajjanapu:
@anantprsd5 that's not an issue. we can show some instructions when the user first time interacts with that button. I've seen this kind of method in some paid camera apps.

mohitmanuja:
@vinaysajjanapu Yes, okay yes we can use good effects later on.
@anantprsd5 We can use Showcase to highlight the instruction when user will install our app.
I think we should include a showcase for the whole app. 
Example : https://github.com/amlcurran/ShowcaseView

anantprsd5:
[...]

PR #624, opened 2017-06-18
Hide toggle button while taking picture and long press to reset color effects

anantprsd5:
Fixes issue #598 

Changes: 
1. Now toggle button hides itself while taking picture.
2. Added long click listener to reset color effect.
3. Removed all color effects list from pop up view.
4. Add showcase view to show toggle button functionality.


app/build.gradle
```diff
@@ -119,4 +119,5 @@ dependencies {
     compile "com.android.support:support-v4:$rootProject.supportLibraryVersion"
     compile "com.facebook.stetho:stetho-okhttp3:$rootProject.stethoVersion"
     compile "com.squareup.picasso:picasso:$rootProject.picassoVersion"
+    compile 'com.github.amlcurran.showcaseview:library:5.4.3'
```
vinaysajjanapu: don't use inline version code.

PR #641, opened 2017-06-20
fix: outOfMemory Exception. Removed the last edited action added latest action in list

### LINKED ISSUE #640: Handle undo in editor when outofmemory exception is caught
**Actual Behaviour**

There is no proper implementation after outofmemory exception is caught.

**Expected Behaviour**

Release memory somehow by either removing previous oldest tracking image and add current tracking image to end.

**Steps to reproduce it**

Open image in editor. keep on editing multiple times ( may be 15) and press undo.

I like to. But if anyone wants to work on it they can.

pa1pal:
Fix  #640 

Changes: 
- [x]  Removed the last edited action and added the latest action in the list
- [x]  change the visibility of undo button from invisible to gone to fix the UI of button
- [x]  Change the PR templace

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/editimage/EditImageActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/editimage/EditImageActivity.java
@@ -355,10 +355,17 @@ public class EditImageActivity extends EditBaseActivity implements View.OnClickL
 
     private void addToUndoList() {
         try{
+            TODO:// implement a more efficient way, like storing only the difference of bitmaps or
+            // steps followed to edit
+
             bitmapsForUndo.add(mainBitmap.copy(mainBitmap.getConfig(),true));
         }catch (OutOfMemoryError error){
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/phimpme/editor/editimage/EditImageActivity.java
```diff
@@ -355,10 +355,17 @@ public void changeMainBitmap(Bitmap newBit) {
...
+
             bitmapsForUndo.add(mainBitmap.copy(mainBitmap.getConfig(),true));
         }catch (OutOfMemoryError error){
-            //Snackbar.make(getWindow().getDecorView().getRootView(),"Out of Memory. steps = " + bitmapsForUndo.size(),Snackbar.LENGTH_LONG).show();
-
+            /**
+             * When outOfMemory exception throws then to make space, remove the last edited step
+             * from list and added the new operation in the end.
+             */
+            bitmapsForUndo.remove(0);
```
vinaysajjanapu: remove the bitmap of index 1 from the list. 
because when we keep on undoing, it would be better if we end up with the original image rather than some randomly edited image.

I am not sure whether only just removing bitmap from the list would clear memory. I think you should call bitmap.recycle before removing that bitmap from the list.

PR #650, opened 2017-06-21
feat: added facebook login in account manager

### LINKED ISSUE #649: Connect facebook in account manager
**Actual Behaviour**

Currently, only Twitter sign in is active on account manager. Will add facebook login.

**Expected Behaviour**

When user select facebook from the dialog fragment in account manager,  facebook login will appear

**Steps to reproduce it**

Open account manager --> add account from toolbar 

Yes

pa1pal:
Fix #649 

Changes: Uses Login manager from facebook SDK to integrate fb login. Added database part in onSuccess of that.


pa1pal:
The code in this PR works fine, but a issue with the CallBackManager. Created issue #653 for the same, need to change the dialog fragment to activity. Will open another PR.

PR #656, opened 2017-06-22
Checking if Instagram is installed or not and handling No internet connection error

### LINKED ISSUE #635: Check if Instagram is installed or not
**Actual Behaviour**

If Instagram is not installed then a white screen appears showing error message "No app is installed to complete the action"

**Expected Behaviour**
Display snackBar to display the message. 

**Steps to reproduce it**
ShareActivity->Instagram upload button

Yes.

Subhankar29:
Fix #635 and #664  

Changes: Display snackBar to display the message if Instagram is not installed. Display Snackbar if internet connection is not available.

- Check if Instagram is installed or not.

- Check if internet connection is available or not.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/SharingActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/SharingActivity.java
@@ -5,10 +5,12 @@ import android.app.AlertDialog;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ActivityInfo;
+import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.ColorDrawable;
+import android.net.ConnectivityManager;
 import android.net.Uri;
 import android.os.Build;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/phimpme/SharingActivity.java
```diff
@@ -272,15 +273,30 @@ private void otherShare() {
...
+            share.putExtra(Intent.EXTRA_STREAM, uri);
+            share.setType("image/*");
+            share.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+            startActivity(Intent.createChooser(share, caption));
+            atleastOneShare = true;
+        }else
+            Snackbar.make(parent,R.string.instagram_not_installed,Snackbar.LENGTH_LONG).show();
+    }
+
+    private boolean isAppInstalled(String packageName) {
```
pa1pal: Move this function in `Utils` class, because its generic and if we require  to use this in future, Don't need to create again.

Subhankar29: Sure will do that.

PR #657, opened 2017-06-22
Description option added in SingleMediaActivity

### LINKED ISSUE #630: Add description option for media.
**Actual Behaviour**
No option exists to allow the user to write a description for each media.

**Expected Behaviour**
An option will exist in SingleMediaActivity.java which will allow the user to write a description for the currently open media. The user can also see previously written description when he opens a media.

Yes.

Nirvan101:
Fix #630 Added a description option in the overflow menu of SingleMediaActivity.java .
Using this the user can add a description (about the media) for each media. So for every media, the user can write a passage describing the it, the story behind the it etc.

Changes:
Added an option 'Description' in the overflow menu of SingleMediaActivity.java. 
Added DescriptionActivity. The user is led here after he clicks on Description option above.
Added a Realm database for saving descriptions of all media. For this I introduced the RealmController class and the Item.class. The latter is the object type that is stored in the Realm table.
Also introduced a few lines in LFMainActivity and SingleMediaActivity to pass the file-path to DescriptionActivity. 

Screenshots:
screenshot_20170623-124312
screenshot_20170623-124322
screenshot_20170623-124333


mohitmanuja:
@Nirvan please post some screenshots.

Nirvan101:
@mohitmanuja screenshots added.

app/src/main/java/org/fossasia/phimpme/leafpic/activities/DescriptionActivity.java
```diff
@@ -0,0 +1,69 @@
...
+    Item temp;
+    String path;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_description);
+
+        //realmController=new RealmController();
+        editText = (EditText) findViewById(R.id.activity_description_editText);
```
the-dagger: https://github.com/fossasia/phimpme-android/pull/660#pullrequestreview-45916091

PR #667, opened 2017-06-23
changed the layout account manager using swithes for sign in

pa1pal:
Fix #665 

Changes: 
- [x] Added switch to sign in
- [x] CardView layout
- [x] set things in recyclerview.

#### Screenshot
account_manager_new

PR #674, opened 2017-06-24
Added Sign In functionality in AccountActivity

### LINKED ISSUE #653: Issue with instance of Callback manager in account manager for various accounts.
**Actual Behaviour**

Currently, we are using Dialog fragment while connecting new account, which is sometimes inefficient, as we are not aware of what type of callback the service will provide. Like in case of facebook, instance of `callBackManager` getting changed in `onActivityResult`.
So I will shift this to new efficient mode of adding accounts on Activity instance only. I research on some similar apps as well. The Switch behavior of signIn would look good. 

**Expected Behaviour**

The account Manager will already have Account Name and a toggle button to sign in. 

**Screenshots of the issue**

image

Yes

pa1pal:
Fix #653  (Should be merge after #667)

Changes: 

- [x] Removed the dialog fragment
- [x] Added sign in in AccountActivity
- [x] Removed the menu item
- [x] Sign In switch toggle after sign in

Screenshot:

image


pa1pal:
@mohitmanuja @vinaysajjanapu @Subhankar29 @anantprsd5 Please review.

PR #701, opened 2017-06-30
Integrated opencv4android into Phimpme.

### LINKED ISSUE #688: Integrate openCV4Android in to the app
**Actual Behaviour**

All the image processing in editor is done using raw functions which are not optimized.

**Expected Behaviour**

OpenCV library has very large number of optimized IP functions. If we integrate it, we can optimize image editor's performance and can implement many other face processing which were very difficult using raw functions.

**Cons**

Adding OpenCV creates lib files for all architectures. By default lib files for all architectures gets included in apk which is unnecessary. So this causes apk size to increase a lot. But if we generate multiple apks (different apk for different architecture) we can use openCV and app size will also be very less. Most apps follow this method. If we upload apks of all architectures to playstore, it takes care of which one to download based on our device config.

So, I think we can integrate OpenCV4Android into our app. 

What say @pa1pal @mohitmanuja @anantprsd5 @Subhankar29 ?

yes.

vinaysajjanapu:
Fix #688 
Changes: 
- [x] Integrated OpenCv native headers, files and module
- [x] Changed dependencies
- [x] Changed build scripts
- [x] Implemented architecture wise split in apks
- [x] Travis Script (@pa1pal)


vinaysajjanapu:
@the-dagger Codacy won't work well with native code.

vinaysajjanapu:
@the-dagger the failing codacy test is because of the function names in imported opencvmodue. It is very difficult to make those changes and I don't think those changes should be made because changing function names in the module affects many other functions.

the-dagger:
@vinaysajjanapu Sure, no worries.

PR #705, opened 2017-07-02
Fixed outOFMemory exception while loading image in SingleMediaActivity

### LINKED ISSUE #689: App crashes sometimes due to OutOfMem. Exception
**Actual Behaviour**

App crashes when clicked on any high resolution images and while trying to upload high quality images to twitter.

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

1. Open any  album.
2. Arrange photos according to size in descending order.
3. Click on the image at 0 index.
4. Picture won't load in single media activity or app will crash.
5. Now again try to upload the same image to twitter.

Yes

anantprsd5:
Fix #689 

Changes: Added large memory heap in the manifest and preferred bitmap RGB565.


app/src/main/AndroidManifest.xml
```diff
@@ -17,6 +17,7 @@
 
     <application
         android:name=".MyApplication"
+        android:largeHeap="true"
```
the-dagger: @anantprsd5 this is just a hack around that lets you app use more memory.
A better approach to this issue is to cache the image instead.

Please check the following link for more info on this, 
https://developer.android.com/topic/performance/graphics/cache-bitmap.html

anantprsd5: @the-dagger The Glide library we are using here should actually take care of these things right? Even in developer page you mentioned it is recommended to use Glide library for most purposes. And for the caching part, I guess we are doing that using the Glide library here.
```
.diskCacheStrategy(DiskCacheStrategy.SOURCE)
```

anantprsd5: @the-dagger Making the app use more memory and using the RGB 565 to load bitmaps solved the crash I was facing on my device.

PR #709, opened 2017-07-02
Added instructions to setup opencv to readme

### LINKED ISSUE #708: Update Readme file with instructions to add OpenCV4Android static libraries
**Actual Behaviour**

Project won't get built if the static library files are not present in proper location in the project.

**Expected Behaviour**

Instructions should be present in Readme file for setting up these static library files of opencv in the project.

yes

vinaysajjanapu:
Fix #708 

Changes:
Modified readme and added instructions.

PS: merge this after PR #701


the-dagger:
@mohitmanuja create a new branch from master and then send a PR from there.

mohitmanuja:
@vinaysajjanapu

pa1pal:
@the-dagger why from master?
The branch in this PR created from #701. When #701 will get merge then I think we can `cherry-pick` and `force push` to solve the commits issue?

the-dagger:
@pa1pal all that overhead isn't really needed if the PR is sent correctly.
A better approach would've been to send the readme commit in #701 itself.

pa1pal:
Yes that will be the better approach.

PR #825, opened 2017-07-18
Added Dropbox image upload functionality

### LINKED ISSUE #791: Implement option to upload images on Dropbox
**Actual Behaviour**

No option to upload images on Dropbox.

**Expected Behaviour**

Dropbox implementation should be there.

Yes.

anantprsd5:
Fix #791 

Changes: Added Dropbox image upload functionality from the Sharing Activity. 

**To upload image**

1. Login from the account manager.
2. Open share activity.
3. Click on Dropbox option

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
@@ -61,6 +61,8 @@ import java.util.List;
 import io.realm.Realm;
 import io.realm.RealmQuery;
 
+import static org.fossasia.phimpme.utilities.Constants.*;
+
 /**
  * Created by pa1pal on 13/6/17.
  */
@@ -87,9 +89,6 @@ public class AccountActivity extends ThemedActivity implements AccountContract.V
     private static final int NEXTCLOUD_REQUEST_CODE = 1;
     private static final int RESULT_OK = 1;
     public static final int IMGUR_KEY_LOGGED_IN = 2;
-
-    final static private String APP_KEY = "APP_KEY";
-    final static private String APP_SECRET = "API_SECRET";
 
     private DropboxAPI<AndroidAuthSession> mDBApi;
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -242,7 +238,7 @@ public void onClick(DialogInterface dialog, int which) {
     }
 
     private void signInDropbox() {
-        if (accountPresenter.checkAlreadyExist(accountName[7]))
+        if (accountPresenter.checkAlreadyExist(accountName[8]))
```
the-dagger: @anantprsd5 Better idea is to store these integers as constants with proper variable names.

app/src/main/java/org/fossasia/phimpme/utilities/Constants.java
```diff
@@ -26,5 +26,8 @@
 
     public static String IMGUR_HEADER_CLIENt = "Client-ID";
     public static String IMGUR_HEADER_USER = "Bearer";
+
+    public final static String APP_KEY = "9qgi9rcomdglx8f";
```
the-dagger: Not sure if exposing this is a good idea.
Maybe we can have an internal testing with this api key and if it works well then we can remove this and replace it with a dummy string.

anantprsd5: @the-dagger Made the required changes, added the description to configure the app keys in the README.md file.

app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -208,7 +207,6 @@ public void onItemClick(final View childView, final int position) {
 
                 case 8:
```
the-dagger: Do this for all the integers @anantprsd5

anantprsd5: @the-dagger Done. Replaced all the constants in the ShareActivity.java file.

app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
pa1pal: @the-dagger Yes using constants is a better idea. 
I think even more better approach is if we use Enumerated Type in the model class itself, that helps us to avoid using indexing of an array all the time, which is highly error prone.

PR #832, opened 2017-07-18
Cleared image cache in Shareactivity

### LINKED ISSUE #824: Image not refreshed in shareActivity
**Actual Behaviour**

In share, activity opens any pic after that delete that pic then renames any pic with the same name which you deleted. The last image is opening due to cache.

**Expected Behaviour**

Clear cache and open fresh image.

Yes.

mohitmanuja:
Fix #824 

Changes: 
Cleared image cache in Shareactivity.


app/google-services.json
```diff
@@ -0,0 +1,42 @@
+{
```
the-dagger: @mohitmanuja Please do not add these keys to the public repo.
If you plan on adding them, create a dummy project on firebase and then add the son from that over here.

mohitmanuja: @the-dagger Thanks for pointing out, I will remove all keys from the repo and will send the separate PR using firebase JSON.

mohitmanuja: @the-dagger Please approve this PR.

PR #849, opened 2017-07-20
login twitter only from account manager

### LINKED ISSUE #807: Error on first time on sharing on twitter
**Actual Behaviour**

Some error occurs when clicking first time on twitter from sharing activity and again showing authorising twitter.

**Expected Behaviour**

It shouldn't occur.

**Steps to reproduce it**

First, Clear app data.
Setup twitter account from the accounts section. Then go to share activity and click on twitter.

**Screenshots of the issue**

@Subhankar29 @pa1pal If you want to work on this you can.

pa1pal:
Fix #807 

Changes: Twitter login only from account manager, removing the authorization from sharing activity.

Work in progress.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
@@ -439,7 +439,8 @@ public class AccountActivity extends ThemedActivity implements AccountContract.V
                     // Writing values in Realm database
                     account.setAccountname(TWITTER);
                     account.setUsername(session.getUserName());
-                    account.setToken(String.valueOf(session.getAuthToken()));
+                    account.setToken(String.valueOf(session.getAuthToken().token));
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/fossasia/phimpme/sharetwitter/HelperMethods.java
```diff
@@ -50,22 +47,40 @@ public static void postToTwitterWithImage(Context context, final String imageUrl
...
-				e.printStackTrace();
-				success = false;
-			}
-		}else{
-			Log.d(TAG, "----- Invalid File ----------");
-			success = false;
-		}
-		postResponse.onFinsihed(success);
 
+		new Thread(new Runnable() {
```
mohitmanuja: You are creating a new thread it is already running in the background thread. Aysync task provides a background thread so no need to create a new thread here.

PR #851, opened 2017-07-20
Added realm in flickr and updated details in accounts page

### LINKED ISSUE #821: Use realm database in Flickr Share
**Actual Behaviour**

Currently the oauth tokens and other details are stored in shared preferences. and the username is not getting updated in accounts page.

**Expected Behaviour**

Store all the details in realm database and make changes in accounts activity for updating with flickr.

**Steps to reproduce it**

Open accounts and login to flickr

yes

vinaysajjanapu:
Fix #821 

Changes: 
- Added realm database in flickr for storing tokens and other details of account.
- Updated details in accounts page
- Added caption support

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
@@ -2,8 +2,6 @@ package org.fossasia.phimpme.share.flickr;
 
 import android.content.Context;
 import android.content.Intent;
-import android.content.SharedPreferences;
-import android.content.SharedPreferences.Editor;
 import android.content.pm.ActivityInfo;
 import android.net.Uri;
 import android.os.Bundle;
@@ -11,6 +9,7 @@ import android.os.Handler;
 import android.widget.Toast;
 
 import org.fossasia.phimpme.base.ThemedActivity;
+import org.fossasia.phimpme.data.local.AccountDatabase;
 import org.fossasia.phimpme.share.flickr.tasks.GetOAuthTokenTask;
 import org.fossasia.phimpme.share.flickr.tasks.OAuthTask;
 import org.fossasia.phimpme.share.flickr.tasks.UploadPhotoTask;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
```diff
@@ -128,52 +127,58 @@ public void onOAuthDone(OAuth result) {
...
-		photoStream = inputStream;
-	}
-
-	public static InputStream getInputStream(){
-		return photoStream;
-	}
-
-	public static void setFilename(String file){
-		filename = file;
+		Realm realm = Realm.getDefaultInstance();
```
pa1pal: Create a global variable for `realm` as you are using in above `getOAuthToken()`as well. Or adding some function which returns the results, so we can remove the redundant code.

app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
```diff
@@ -128,52 +127,58 @@ public void onOAuthDone(OAuth result) {
...
+		AccountDatabase account;
+		realm.beginTransaction();
+		if( result.size() == 0 ) {
+			account = realm.createObject(AccountDatabase.class,FLICKR.toString());
+			account.setToken(token);
+			account.setTokenSecret(tokenSecret);
+			account.setUsername(userName);
+			account.setUserId(userId);
+		} else {
+			account = result.first();
```
pa1pal: Didn't get this part. What I understand is that you are checking if user is logged in then update its values. But currently we support at max one user per account and following the account manager, so User can log out and re login again. No update of values.

app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
vinaysajjanapu: The two places where realm is used are in separate threads. If I use same object for both threads, it is causing a crash.

app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
vinaysajjanapu: What you understood is wrong. I accept that I didnot write anything for logging out. but here it is completely for logging in. In the first call to db secret token gets saved and other elements will be empty. In the second call other details gets updated.

vinaysajjanapu: Logging out part has to be written in accountactivity, not here. I'll do that.

pa1pal: What has already setup in Account Manager, you don't need to add code for logging out. Because once an entry added in to the DB the switch of that account turned on. There is no such feature of updating the current values. Because once you click on **ON** switch. It shows dialog to signout, which basically deleting the entry from database.
I am not getting the why there is else block, when we create object everytime.

> In the first call to db secret token gets saved and other elements will be empty. In the second call other details gets updated.

If this is the case then, only add those lines which are necessary in that block.

app/src/main/java/org/fossasia/phimpme/share/flickr/FlickrActivity.java
pa1pal: What is the error?

mohitmanuja:
@vinaysajjanapu Please resolve conflicts.

mohitmanuja:
@vinaysajjanapu Flickr is working perfectly. 

Needs minor changes i.) Use the material design progress bar to show authenticating and for uploading.
ii.)  After successful uploading show, snack bar uploaded successfully. 
You can do this changes in a different issue.

PR #864, opened 2017-07-24
Added description and URL in the Pinterest dialog box

### LINKED ISSUE #812: Adding instruction in a Dialog Box to get the Board ID while uploading Image on Pinterest
**Actual Behaviour**

Currently, there is no instruction to add the Board ID in the Dialog Box

**Expected Behaviour**

To add instruction about how to get Board ID, in the dialog box. 

**Steps to reproduce it**

SharingActivity-> Pinterest

Yes.

Subhankar29:
Fix #812 

Changes: 
Changed the dialog box style to the Material design. 
Added URL to get the Board ID

Steps to reproduce:-> SharingActivity->Pinterest

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/leafpic/util/AlertDialogsHelper.java
+++ b/app/src/main/java/org/fossasia/phimpme/leafpic/util/AlertDialogsHelper.java
@@ -1,12 +1,15 @@
 package org.fossasia.phimpme.leafpic.util;
 
 import android.content.Intent;
+import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.Typeface;
 import android.net.Uri;
 import android.support.annotation.StringRes;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.CardView;
+import android.text.Html;
+import android.text.method.LinkMovementMethod;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/fossasia/phimpme/leafpic/util/AlertDialogsHelper.java
```diff
@@ -61,6 +64,35 @@ public static AlertDialog getInsertTextDialog(final ThemedActivity activity, Ale
         dialogBuilder.setView(dialogLayout);
         return dialogBuilder.create();
     }
+    public static AlertDialog getInsertTextDialogLink(final ThemedActivity activity, AlertDialog.Builder dialogBuilder , EditText editText, String link) {
```
vinaysajjanapu: @Subhankar29, Why don't you use this function from AlertDialogHelper.java class(https://github.com/fossasia/phimpme-android/blob/development/app/src/main/java/org/fossasia/phimpme/leafpic/util/AlertDialogsHelper.java#L37).
It is a static function and you can call it from anywhere. Why are you creating the same function again? Please avoid creating duplicate functions.

Subhankar29: @vinaysajjanapu Yes, I am aware of that. Since I wanted to create an alert dialog box with a clickable link, it wouldn't be possible in the existing AlertDialogBox.

vinaysajjanapu: You can modify the existing function as per your requirement. no need to create a new function again

Subhankar29:
@vinaysajjanapu Made changes as suggested.

mohitmanuja:
@Subhankar29  Delete pinterest_dialog.xml

Subhankar29:
@mohitmanuja Done, Please review.

PR #898, opened 2017-07-28
added check for server address protocol in NextCloud and OwnCloud

### LINKED ISSUE #895: NextCloud and OwnCloud server url causing failed to upload image
**Actual Behaviour**

Nextcloud and OwnCloud both shares the same android-library. The server url user entered is causing failed to upload the image.

**Expected Behaviour**

Server url entry into the database Adding the protocol (http:// or https://) will fix the error. 
adding Url Input check.

Yes

pa1pal:
Fix #895 

Changes: Added check to add server address protocol in OwnCloud and NextCloud login page. Image wasn't uploading due to this. Tested and uploading image working.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
@@ -324,7 +324,7 @@ public class SharingActivity extends ThemedActivity implements View.OnClickListe
                 break;
 
             case NEXTCLOUD:
-                shareToNextCloud(1);
+                shareToNextCloudAndOwnCloud(getString(R.string.nextcloud));
                 break;
 
             case PINTEREST:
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -320,7 +320,7 @@ public void onItemClick(View childView, int position) {
                 break;
 
             case NEXTCLOUD:
-                shareToNextCloud(1);
+                shareToNextCloudAndOwnCloud(1);
```
mohitmanuja: Use Constant Int variable instead of 1/2.

pa1pal: @mohitmanuja done.
Used string as parameter, Int wasn't making any sense.

PR #899, opened 2017-07-28
shows only connected accounts in share activity

### LINKED ISSUE #845: Show only connected accounts in sharing activity
**Actual Behaviour**

All the accounts are visible in list whether they are connected or not.

**Expected Behaviour**

It would be better if we show only connected accounts in sharing activity

**Steps to reproduce it**

Select an image -> click share

yes

vinaysajjanapu:
Fix #845 
Changes:

Shows only the accounts that are connected from account activity
Shows imgur always and instagram, whatsapp if installed


mohitmanuja:
@vinaysajjanapu Also add WhatsApp. Wait for #903 to merge.

vinaysajjanapu:
@mohitmanuja Added Whatsapp. Please review.

PR #910, opened 2017-07-30
Google plus & Instagram handled correctly in sharing activity

### LINKED ISSUE #894: Google plus authentication is not working
**Actual Behaviour**

Currently, I am not able to sign in and upload image to googleplus.

**Expected Behaviour**

It should work.

**Steps to reproduce it**

1. Open accounts.
2. Tab on google plus.
3. Select an account.
You will get a message login fail.

**Screenshots of the issue**

Yes

mohitmanuja:
Fix #894 

Changes:

- [x] Removed google plus authentication.
- [x] Added a New variable in account database to handle hidden number.
- [x] Check if google plus is installed then it will visible in a share account.

Because we cannot google plus api. The only way to use google plus domain API and it is paid.


mohitmanuja:
@vinaysajjanapu I have added telegram in this PR.

mohitmanuja:
@Subhankar29 @pa1pal @vinaysajjanapu @anantprsd5  review this PR.

PR #973, opened 2017-08-07
Fix #912 - Set color of tab icon bottom navigation view

### LINKED ISSUE #912: Background Color of Tab Icon Should Adjust According to Primary Colour
**Actual Behaviour**

The background colour for highlighting the current tab is fixed. So if the primary colour is same as the highlight colour the tab icon becomes invisible. 

**Expected Behaviour**
The background colour of the icon should adjust according to primary colour. 

**Steps to reproduce it**
Switch primary colour to dark grey
**Screenshots of the issue**
image

Yes

mayank8318:
Fix #912 

Screenshots for the change: 
image
image

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/base/BaseActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/base/BaseActivity.java
@@ -1,9 +1,12 @@
 package org.fossasia.phimpme.base;
 
 import android.content.Intent;
+import android.content.res.ColorStateList;
+import android.graphics.Color;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.design.widget.BottomNavigationView;
+import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/phimpme/base/BaseActivity.java
```diff
@@ -18,11 +20,25 @@
...
+    private int[] colors = new int[] {
+            Color.WHITE, // checked
+            0 // unchecked set default in onCreate
+    };
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(getContentViewId());
+        colors[1] = getResources().getColor(R.color.bottom_navigation_tabs , getTheme());
```
vinaysajjanapu: try using this once
    colors[1]  = ContextCompat.getColor(this, R.color.bottom_navigation_tabs);

PR #974, opened 2017-08-07
modified the zoom controls using vector assets

### LINKED ISSUE #970: Redesign the exposure control view in CameraActivity
**Actual Behaviour**

The exposure control view is looks quite old. 

**Expected Behaviour**

Redesign it with material components.

**Steps to reproduce it**

Open Camera --> Exposure Compensation

**Screenshots of the issue**
screen shot 2017-08-07 at 12 08 02 am

yes

pa1pal:
Fix #970 

Changes: Removed the `zoom control` and added the Image Buttons.

Screenshots for the change: 
screen shot 2017-08-07 at 8 40 57 pm

It is showing large changes because I re formatted one file


pa1pal:
@vinaysajjanapu I agree with you, It is not necessary, but not useless as well. It's better to provide both the features, because it depend on User preference. In the context of camera, I think these buttons are much useful than seekbar. Try using it once.

PR #1020, opened 2017-08-13
Changed edited image format to that of input image

### LINKED ISSUE #993: Editing a transparent png is returning a normal jpg
**Actual Behaviour**

Transparacy is not maintained for a png image

**Expected Behaviour**

If an image has an alpha channel, it should be maintained.

**Steps to reproduce it**

Edit a transparent png image in editor. 

yes.

vinaysajjanapu:
Fix #993 
Changes: 
found the extension from the path of the image and created the output file with that extension

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/FileUtils.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/FileUtils.java
@@ -43,12 +43,23 @@ public class FileUtils {
 		return Environment.getExternalStorageDirectory();
 	}
 
-	public static File genEditFile(){
+	public static File genEditFile(String extension){
 		return FileUtils.getEmptyFile("phimpme"
-				+ System.currentTimeMillis() + ".jpg");
+				+ System.currentTimeMillis() + extension);
 	}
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/phimpme/editor/FileUtils.java
```diff
@@ -43,12 +43,26 @@ public static File createFolders() {
...
-	public static File genEditFile(){
+	public static File genEditFile(String extension){
 		return FileUtils.getEmptyFile("phimpme"
-				+ System.currentTimeMillis() + ".jpg");
+				+ System.currentTimeMillis() + extension);
 	}
 
-	public static File getEmptyFile(String name) {
+	public static String getExtension(String path){
+		String lc = path.toLowerCase();
```
anantprsd5: It can be done like this
return lc.substring(lc.lastIndexOf("."))

vinaysajjanapu: Sure @anantprsd5.
Made changes. please review. thanks.

PR #1082, opened 2017-08-26
Removed the crashlytics from development branch

### LINKED ISSUE #1081: Remove crash reports from the development branch
**Actual Behaviour**

Currently, the crash reports are sent from the app installed from the development branch as well.

**Expected Behaviour**

Remove the crash reports from the development branch.

Yes

anantprsd5:
Fix #1081 

Changes: Removed Crashlytics from the development branch


mohitmanuja:
@anantprsd5 please use a boolean variable isPublish which we will set true only in master branch. And if it is true then only initialize fabric. Don't remove crashyltics because when it gets merged in master branch it will completely removed.

PR #1128, opened 2017-09-13
Navigation Drawer closes smoothly

### LINKED ISSUE #1124: Lag while closing of Navigation Drawer
**Actual Behaviour**
Currently when we select either of **Settings**,**Upload History**,**About** from the Navigation Drawer the intent for the new Activity is fired and due to this the exit animation of the drawer isn't completed.Because of this the animation doesn't appear to be smooth.

**Expected Behaviour**
The exit animation should be completed fully and after that a new intent should be fired.

**Steps to reproduce it**
In the navigation drawer select either of the options mentioned above.

Yes!

@mohitmanuja Can I work on this issue?

sridharjajoo:
Fix #1124 
Changes: 
When the user selects either of **Settings** , **About** , **Upload History** from the navigation drawer , the drawer first closes **smoothly** and then an intent gets fired which further starts a new activity corresponding to the option selected.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -16,6 +16,7 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.Handler;
 import android.provider.MediaStore;
 import android.support.v4.content.ContextCompat;
 import android.support.v4.view.GravityCompat;
@@ -70,6 +71,7 @@ import org.fossasia.phimpme.gallery.util.StringUtils;
 import org.fossasia.phimpme.gallery.views.GridSpacingItemDecoration;
 import org.fossasia.phimpme.uploadhistory.UploadHistory;
 import org.fossasia.phimpme.utilities.ActivitySwitchHelper;
+import org.fossasia.phimpme.utilities.Constants;
 import org.fossasia.phimpme.utilities.SnackBarHandler;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


sridharjajoo:
@anantprsd5 @mohitmanuja Kindly review the P.R
Thank You

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -651,27 +653,42 @@ private void setDrawerTheme() {
...
-                Intent intent = new Intent(LFMainActivity.this, SettingsActivity.class);
                 mDrawerLayout.closeDrawer(GravityCompat.START);
-                startActivity(intent);
+                new Handler().postDelayed(new Runnable() {
+                    @Override
+                    public void run() {
+                        Intent intent = new Intent(LFMainActivity.this, SettingsActivity.class);
+                        startActivity(intent);
+                    }
+                }, Constants.NAVIGATION_DRAWER_EXIT_TIME);
```
vinaysajjanapu: @sridharjajoo Instead of using a handler with hard-coded drawer exit time, why don't you use a drawer listener?

app/src/main/java/org/fossasia/phimpme/utilities/Constants.java
```diff
@@ -46,6 +46,8 @@
     public final static String PACKAGE_PLAYSTORE = "com.android.vending";
     public final static String PACKAGE_TELEGRAM = "org.telegram.messenger";
 
+    public final static int NAVIGATION_DRAWER_EXIT_TIME = 260;
```
vinaysajjanapu: see above comment

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
sridharjajoo: I initially used onDrawerClosed listener for implementing the solutoion but it is only available for API
level 22 and higher.So in cases where API level is less than 22 a lag will still be there.

sridharjajoo: Should I send a PR which uses drawer listener?

vinaysajjanapu: @sridharjajoo yes, please.

sridharjajoo: @vinaysajjanapu 
I am sending a new PR . Kindly close this.
Thank You

PR #1188, opened 2017-09-21
Fixes #1187 Warning on pressing back button in all edit modes

### LINKED ISSUE #1187: Warning on pressing back button in all edit modes
**Actual Behaviour**

In any editing mode, performing either of the functions (adding stickers, enhancing, etc.) and pressing back button without saving leads to all changes being discarded.

**Expected Behaviour**

A prompt should be displayed asking if the user wants to discard all the changes or continue editing.

**Steps to reproduce it**

Open the app, choose any image and add any sticker. Then press the back button. All stickers added will be removed.
Similar observation can be made on editing it in any mode.

**Screenshots of the issue**

img_20170921_044141

img_20170921_044116

Yes

nb2998:
Fix #1187 

Changes: A warning mesage displayed asking if the user wants to discard all the changes or continue editing.

Screenshots for the change: 

img_20170921_044141

img_20170921_044116


vinaysajjanapu:
I've also seen that you are using the common message in all dialogs. Please make it conditional.

nb2998:
@vinaysajjanapu Okay.. And is conditional message like "Do you want to discard all the stickers?" fine?

nb2998:
@vinaysajjanapu  I just noticed that rotation seekbar changes are persisted even if user chooses to discard. So, when the image is rotated and saved more than once, the previous progress of seekbar gets added to the current orientation of image.
One way to solve this is setProgress to 0 in onPause() of RotateFragment which again has some limitations. So, is there a better way to solve this?

PR #1193, opened 2017-09-22
Fix #1167 -Highlight the sub-edit option when selected.

### LINKED ISSUE #1167: Highlight edit sub-option just like the edit option
**Actual Behaviour**

The edit sub-option like HUE in the screenshot is not highlighted when it is selected.

**Expected Behaviour**

It should be highlighted

**Steps to reproduce it**

Go to edit any photo and select any effect to add.

**Screenshots of the issue**

21298069_1806911089337697_5885298401859338240_n

Yes
@mohitmanuja @vinaysajjanapu Should this be there?

mayank8318:
Fix #1167 

Changes: Highlight the effect under consideration.

Screenshots for the change: 
18946638_1812062235489249_2006984994186067968_n

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/fragment/RecyclerMenuFragment.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/fragment/RecyclerMenuFragment.java
@@ -3,9 +3,11 @@ package org.fossasia.phimpme.editor.fragment;
 import android.content.Context;
 import android.content.res.TypedArray;
 import android.graphics.Bitmap;
+import android.graphics.Color;
 import android.graphics.Matrix;
 import android.os.AsyncTask;
 import android.os.Bundle;
+import android.support.v4.content.ContextCompat;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/fossasia/phimpme/editor/fragment/RecyclerMenuFragment.java
```diff
@@ -205,15 +219,35 @@ public void onBindViewHolder(mRecyclerAdapter.mViewHolder holder, final int posi
...
                     itemClicked(position,v);
                 }
             });
         }
 
+        private void highlightSelectedOption(int position, View v) {
+            currentSelection = position;
+            int color = ContextCompat.getColor(v.getContext(), R.color.md_grey_200);
+            int length = titlelist.length();
+            for(int i = 0;i < length;i++)
```
anantprsd5: Why is for loop used here? It can be done using the position passed in as the parameter.

mayank8318:
@anantprsd5 Actually I was using it to set all the background to Color.TRANSPARENT earlier when I was not storing the currentSelection position.
But now that loop is not required as I already have the index of the last active option.
Thanks :)

PR #1200, opened 2017-09-24
Fix: #1143 Navigation bar color issue

### LINKED ISSUE #1143: Navigation bar color issue
**Actual Behaviour**
App starts with a colored navigation bar, though colored navigation bar option is disabled in the settings.

**Expected Behaviour**
App should start with a black navigation bar as colored navigation bar option is disabled by default in settings.

**Steps to reproduce it**
Opens the app and see the navigation bar color and go to settings you will see colored navigation bar option is disable.

**LogCat for the issue**
.

### Although in settings colored nav bar option is disabled by default

YES!

wookupmaker:
**Fix #1143**

**Changes:**
*Modifies a simple boolean logic  
Now app starts with colored nav bar and in settings, colored-nav-bar option is also enabled.*  

**Screenshots for the change:**

>Nav color is blue and option is also enabled by default when app start for the very first time
  

>Disabling nav bar color option still works fine


wookupmaker:
@mohitmanuja  @anantprsd5  please take a look...

PR #1224, opened 2017-09-27
Fixes #1222 Bottom navbar hides when albums or photos are selected

### LINKED ISSUE #1222: BottomNavigationView should be hidden when albums/photos are selected
**Actual Behaviour**

Currently when the albums or photos are long pressed they get selected but the BottomNavigationView is still visible.

**Expected Behaviour**

BottomNavigationView is not required when the user has selected albums or photos. It should be hidden as long as the user is selecting folders and then again become visible when the user has come out of the editmode.

**Steps to reproduce it**

open the app and long press any album or photo

Yes

nikit19:
Fix #1222 

Changes: BottomNavigationBar is hidden when the user selects albums or photos and becomes again visible when the user presses the back button ,deselects all the selected media or clicks on any menu item.

Screenshots for the change: 

ezgif com-video-to-gif 7

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -112,7 +112,7 @@ public class LFMainActivity extends SharedMediaActivity {
     private Toolbar toolbar;
     private SelectAlbumBottomSheet bottomSheetDialogFragment;
     private SwipeRefreshLayout swipeRefreshLayout;
-    private boolean hidden = false, pickMode = false, editMode = false, albumsMode = true, firstLaunch = true,localFolder=true;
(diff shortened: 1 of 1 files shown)


nikit19:
@anantprsd5 @mohitmanuja  @vinaysajjanapu please review.
thank you

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -169,6 +171,11 @@ private int toggleSelectPhoto(Media m) {
                 selectedMedias.remove(m);
         }
         if (selectedMedias.size() == 0) {
+            if(editMode && hidenav )
+            {
+                getNavigationBar();
+                hidenav=false;
```
anantprsd5: Put this line in the getNavigationBar function

nikit19:
@anantprsd5  is it fine now?

PR #1230, opened 2017-09-28
Fixes #1228 create new folder will not make folders with no name

### LINKED ISSUE #1228: Create new folder makes folders with no name
**Actual Behaviour**

When we click on create new folder and press the spacebar it creates a folder with no name

**Expected Behaviour**

Folders with no name shouldn't be created. There should be a Toast message instead of creating folders with no name.
**Steps to reproduce it**

Long press any album or photo.
Then select move to or copy to.
Click on the icon on  top of the dialog.
Then click on create new folder.
Press the spacebar and click on OK

Yes

nikit19:
Fix #1228 

Changes: Folders with no name or folders which already exist will not be created instead the user will get a toast message
Screenshots for the change: 
ezgif com-video-to-gif 8

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/SelectAlbumBottomSheet.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/SelectAlbumBottomSheet.java
@@ -22,6 +22,7 @@ import android.view.ViewGroup;
 import android.widget.EditText;
 import android.widget.LinearLayout;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import com.mikepenz.community_material_typeface_library.CommunityMaterial;
 import com.mikepenz.iconics.view.IconicsImageView;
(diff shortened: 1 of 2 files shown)


nikit19:
@nb2998  okay I'll update the PR to check that too

nikit19:
ezgif com-video-to-gif 9

app/src/main/java/org/fossasia/phimpme/gallery/SelectAlbumBottomSheet.java
```diff
@@ -128,9 +129,24 @@ public void onClick(View view) {
...
+                  filePosition++;
+                  if (editText.getText().toString().equals(f.getName()))
+                      check=0;
+              }
+              if(!editText.getText().toString().trim().isEmpty() && check!=0) {
 			File folderPath = new File(currentFolderPath.getText().toString() + File.separator + editText.getText().toString());
 			if (folderPath.mkdir()) displayContentFolder(folderPath);
+              }
+              else if(check==0)
+                  Toast.makeText(getContext(),R.string.folder_name_exists,Toast.LENGTH_SHORT).show();
```
anantprsd5: Use a Snackbar instead

nikit19: @anantprsd5 I think since this is a dialog a toast message would be better because when I use a snackbar it is not visible till I move the dialog to full screen. What do you think?
ezgif com-video-to-gif 10

PR #1276, opened 2017-10-03
Fix #1258 bug with onResume method call

### LINKED ISSUE #1258: Bug when activity goes to onPause and back to onResume
**Actual Behaviour**

If any image/album is selected and the user leaves the app(for eg, goes back to home screen), when the user comes back to our app, the state of the app gets changed, i.e all selections are lost

**Expected Behaviour**

State of the app should not change

**Steps to reproduce it**

Select any album, navigate to another app, then come back to the app, all selection will be lost.

**Screenshots of the issue**

giphy 1

Yes!

aditya98ak:
Fix #1258 

Changes: If user goes to another app, while he has selected some images, after coming back, changes should be saved, so there is no need to clear selected image(s)/album(s)

Screenshots for the change: 
giphy


anantprsd5:
@aditya98ak  Remove the commented code

aditya98ak:
@anantprsd5 Changes done!

aditya98ak:
@anantprsd5 @mohitmanuja merge this

PR #1300, opened 2017-10-08
Fix #1265 Paint doesn't cross the photo boundary

### LINKED ISSUE #1265: Canvas size too large while photo editing
**Actual Behaviour**
Currently while we are painting on the photo the strokes can also be made outside the actual photo as shown in the figure below.
image 5
The photo area is only the middle part while the rest of black part isn't a part of the photo.
But still we are able to draw on the black part as well which should't be the case as ultimately when the image is saved , it excludes that particular part as shown below.
image 10

**Expected Behaviour**
The painting strokes should only be made on the photo part and not on the outside.

Yes!
@mohitmanuja @anantprsd5 Can i work on it?

sridharjajoo:
Fix #1265

Changes:
The canvas size which was too large earlier is now clipped to the photo on which editing is to be done.

Screenshots for the change: 
As shown in the image shown below , it is still in editing mode and the paint doesn't go outside the image.

@mohitmanuja @anantprsd5 Kindly review the P.R Thankyou! 
image 19


sridharjajoo:
@mohitmanuja @anantprsd5 Kindly review this PR.
Thank You!

PR #1356, opened 2017-10-17
Fix #1352 use snackbar instead of Toast message

### LINKED ISSUE #1352: "Press back again to exit" should be a SnackBar message
**Actual Behaviour**

Currently a Toast message is displayed when the user tries to exit the app.

**Expected Behaviour**

We should use a SnackBar message instead of Toast message inLFMain activity

**Steps to reproduce it**

Open the app and press the back button

**Screenshots of the issue**

screenshot_20171016-191651

Yes

nikit19:
Fix #1352 

Changes: Used snackbar instead of a Toast message

Screenshots for the change: 
screenshot_20171017-210213


mohitmanuja:
It will be good if you show the exit button instead of default ok. Please show exit and on press of exit, make action to exit the app.

nikit19:
@mohitmanuja changes done
screenshot_20171021-162932

sridharjajoo:
@mohitmanuja @nikit19 @anantprsd5 Just a small suggestion. I think that the current way of exiting the app is good and there is no need of using a snackbar and making it more complicated. What do you guys think?
Thank You!

nikit19:
@sridharjajoo The reason why I thought a snackbar should be used instead of a toast message is that there will consistency. Because almost everywhere a snackBar message is being used in the app. Plus I guess it looks better

sridharjajoo:
@nikit19 Ok! But I have seen the use of Toast message in a lot of apps and so I thought that toast messages are a better option.

Issue #1345, opened 2017-10-14
Improving UI for zooming of picture


aditya98ak:
**Actual Behaviour**

Image is zoomed perfectly but the layout bounds for image view needs to be fixed, Currently, Image is not taking space of status bar and bottom bar when zoomed

**Expected Behaviour**

This should not happen

**Steps to reproduce it**

Zoom any image. It will not make use of the full screen, in fact, its layout will be limited.

**Screenshots of the issue**
This screenshot is of the full zoomed image, still, it is not using space of status bar and bottom bar.
![temp](https://user-images.githubusercontent.com/21143936/31572936-8f6d0bc6-b0ce-11e7-97ac-7c64fa17264e.jpeg)

**Would you like to work on the issue?**

Yes

abishekvashok:
Already fixed via #1360

PR #1360, opened 2017-10-18
Fix #1358 full image is now displayed

### LINKED ISSUE #1358: Full image is not visible in singleMediaActivity
**Actual Behaviour**

Currently bottom part of the image is not visible because the bottom ActionMenuView overlaps it.

**Expected Behaviour**

Bottom ActionMenuView should be translucent so that the full image is visible to the user.

**Steps to reproduce it**

Open any image

**Screenshots of the issue**
screenshot_20171018-004212

Yes

nikit19:
Fix #1358 

Changes: full image is now displayed and the bottom actionMenuView is translucent

Screenshots for the change: 
ezgif com-video-to-gif 5


aditya98ak:
@nikit19 Do this solves [#1345 ](https://github.com/fossasia/phimpme-android/issues/1345)too?
If not, I'll send the PR for #1345

nikit19:
@aditya98ak  Nope. I don't think this PR will solve #1345 too

PR #1272, opened 2017-10-03
Fix #1115 -Dropbox working

codedsun:
Fix #1115 Dropbox sharing with other functionalities working.

Changes: Integrated CloudRail API Service.

Screenshots for the change:

Issue #1115, opened 2017-09-12
App Crashes by Clicking on Accounts Button


codedsun:
**Actual Behaviour**

The App Crashes on Clicking Accounts Buttns

**Expected Behaviour**

The App should work fine

**Steps to reproduce it**

Open Phimpme and click on Accounts Button

**LogCat for the issue**

09-12 22:18:00.212 427-427/org.fossasia.phimpme D/AndroidRuntime: Shutting down VM
09-12 22:18:00.213 427-427/org.fossasia.phimpme E/AndroidRuntime: FATAL EXCEPTION: main
                                                                  Process: org.fossasia.phimpme, PID: 427
                                                                  java.lang.NoClassDefFoundError: Failed resolution of: Lcom/dropbox/client2/session/AppKeyPair;
                                                                      at org.fossasia.phimpme.accounts.AccountActivity.setupDropBox(AccountActivity.java:159)
                                                                      at org.fossasia.phimpme.accounts.AccountActivity.onCreate(AccountActivity.java:153)
                                                                      at android.app.Activity.performCreate(Activity.java:5953)
                                                                      at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1128)
                                                                      at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2267)
                                                                      at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2388)
                                                                      at android.app.ActivityThread.access$800(ActivityThread.java:148)
                                                                      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1292)
                                                                      at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                      at android.os.Looper.loop(Looper.java:135)
                                                                      at android.app.ActivityThread.main(ActivityThread.java:5312)
                                                                      at java.lang.reflect.Method.invoke(Native Method)
                                                                      at java.lang.reflect.Method.invoke(Method.java:372)
                                                                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:901)
                                                                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:696)
[...]

PR #1377, opened 2017-10-20
Fixed crashing of app when DropBox option is selected

### LINKED ISSUE #1368: App crashes when selecting DropBox option in Accounts Activity.
**Actual Behaviour**

The app crashes while selecting DropBox option in Accounts activity and also after the crash when the app restarts, on the bottom menu, Accounts option is selected but gallery has been shown.

**Expected Behaviour**

The app shouldn't crash while selecting DropBox option.

**Screenshots of the issue**

errorimage

Yes, I would like to work on this issue.

raj10071997:
Fix #1368

Changes: Wrong key was entered in manifest , so changed it and entered the correct key . This fixed the crashing of the app issue.


codedsun:
I think the dropbox sdk has been deprciated and they have shifted to v2 ! 
I have made the changes for the issue #1115  by #1272 (Cloudrail is setup !) 

I think we should close this.

PR #1439, opened 2017-10-28
Fix #1432-Added cancel button in the create folder dialog

### LINKED ISSUE #1432: Cancel button not there in the create folder dialog.
**Actual Behaviour**

When the user tries to create a new folder while performing the copy or move operation there is no cancel button to abort the operation, rather the user has to click on the ok button with the edittext being empty to abort.

**Expected Behaviour**

There should be a cancel button to abort the operation.

**Steps to reproduce it**

Try creating a new folder while performing copy or move operation.

Yes.

sauravvishal8797:
Fix #1432 

Changes: Cancel button is added to the dialog. Now on opening any of the dialoges the Ok button will appear disabled at first and will be enabled only if the user updates the text in the edittext.


mayank8318:
@sauravvishal8797 Why remove the button text colour?

sauravvishal8797:
@mayank8318 Yeah. Did the changes. Thanks for mentioning.

PR #1452, opened 2017-11-01
Fix #1371 - Pinning or Unpinning album clears selection mode.

### LINKED ISSUE #1371: On pinning album to top it is still shown as selected
**Actual Behaviour**

When an album is pinned to the top, it is still shown as selected but the selection mode is off.

**Expected Behaviour**

The album should be unselected or selection mode should be on.

**Steps to reproduce it**

Select an album.
Select Pin to Top from overflow menu.

**Screenshots of the issue**

22286920_1840910455937760_6074409819106705408_n

Yes

mayank8318:
Fix #1371 

Changes: Clear selection when album is pinned or un-pinned

Screenshots for the change: 
22551335_1853348788027260_5504657947453554688_n


mayank8318:
@nikit19 @codedsun @aditya98ak @mohitmanuja @anantprsd5 Please review :)

PR #1479, opened 2017-11-11
Fix #1462 - Slideshow bug fixed

### LINKED ISSUE #1462: Slideshow not stopping on click
**Actual Behaviour**
A major bug was that when user enter's the slideshow duration to 1, it will be hard to stop the slideshow by tapping on the screen, and when user enter's the duration to 0, the screen will go black.

**Expected Behaviour**

Slideshow should be stopped on tapping the screen and screen should not go black

**Steps to reproduce it**

1.Enter the slideshow
2.Tap on the screen

Yes!

codedsun:
Fix #1462 Fixed slideshow major bug which I encountered with.

Changes: Added conditions and checks to recover from it.

Screenshots for the change: 
Toast is displayed for slideshow not possible less than 2 sec.

@aditya98ak @mayank8318 @anantprsd5 @mohitmanuja Kindly review!

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
@@ -51,6 +51,7 @@ import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import com.bumptech.glide.Glide;
 import com.mikepenz.community_material_typeface_library.CommunityMaterial;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
```diff
@@ -1153,8 +1154,13 @@ public void onClick(DialogInterface dialog, int which) {
                     slideshow=true;
                     int intValue = Integer.parseInt(value);
                     SLIDE_SHOW_INTERVAL = intValue * 1000;
-                    hideSystemUI();
-                    handler.postDelayed(slideShowRunnable, SLIDE_SHOW_INTERVAL);
+                    if(SLIDE_SHOW_INTERVAL > 1000) {
```
mayank8318: shouldn't this be > 2000 if you are saying 2 sec is the minimum duration?

codedsun: @mayank8318 as the user is allowed only to enter integer values hence by going with 2 sec minimum then it should be >=2000, not >2000 as it will not work for 2 sec also.

codedsun: Hence I have added for the greater than 1 sec :+1:

codedsun:
@mohitmanuja @anantprsd5 Merge this!

app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
mayank8318: oh yes didn't see that it didn't have '=' there.

PR #1557, opened 2017-12-28
Images are now shown properly in landscape and portrait mode

### LINKED ISSUE #1144: Image Not Viewing Properly while toggling between Landscape and Portrait
**Actual Behaviour**

**Expected Behaviour**

**Steps to reproduce it**

**LogCat for the issue**

**Screenshots of the issue**

Yes!

batbrain7:
Fix #1144 

Changes: 

Screenshots for the change: 
screenshot_20171228-171826
screenshot_20171228-171850


batbrain7:
@mohitmanuja please review!!

anantprsd5:
@batbrain7 The PR is too big and I can see that many auto builds are also included in this PR which should have been generated at the time of building the project. Also, I don't think the module has been added in a proper way which is the reason why the diff is so huge. Please close this PR and find another way out. Thanks.

PR #1570, opened 2017-12-31
Fix #1568-All photos delete issue fixed.

### LINKED ISSUE #1568: Issue on performing delete action in all photos section.
**Actual Behaviour**

The updated list is not displayed in the same order as it was before performing the delete action.

**Expected Behaviour**

The updated list should be displayed in the same order as it was before performing the delete action.

**Steps to reproduce it**

Delete photos from all photos section.

Yes.

sauravvishal8797:
Fix #1568 

Changes: The updated list is displayed in the same order after performing delete action as it was before.

### DIFF
--- a/app/src/main/res/layout/dialog_album_detail.xml
+++ b/app/src/main/res/layout/dialog_album_detail.xml
@@ -244,7 +244,7 @@
                       android:id="@+id/album_details_hidden"
                       android:textSize="17dp"
                       android:textColor="@color/icongrey"
-                      android:layout_marginLeft="55dp"/>
+                      android:layout_marginLeft="56dp"/>
           </LinearLayout>
           <View android:layout_width="match_parent"
                 android:layout_height="1dp"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/dialog_album_detail.xml
```diff
@@ -244,7 +244,7 @@
                       android:id="@+id/album_details_hidden"
                       android:textSize="17dp"
                       android:textColor="@color/icongrey"
-                      android:layout_marginLeft="55dp"/>
+                      android:layout_marginLeft="56dp"/>
```
anantprsd5: Can you please tell me why you made changes in the layout?

sauravvishal8797: @anantprsd5 The alignment of the textview here wasn't proper, it required a space indentation.

anantprsd5: @sauravvishal8797 I would suggest you to solve different bugs in separate issues. Fine for now. Thanks.

PR #1578, opened 2018-01-03
Toolbar doesn't expand when soft keyboard appears

### LINKED ISSUE #1201: Background when input is prompted
**Actual Behaviour**

The titlebar expands in the background when any kind of input is prompted from the user.

**Expected Behaviour**

This should not happen.

**Steps to reproduce it**

Try to rename an album, or apply a password.

**Screenshots of the issue**

image

Yes

sridharjajoo:
Fix #1201
Earlier when input was prompted the toolbar would expand.
Now the toolbar doesn't expand.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -184,6 +184,7 @@
         <activity
             android:name=".gallery.activities.LFMainActivity"
             android:configChanges="orientation|screenSize"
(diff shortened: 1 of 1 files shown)


sridharjajoo:
@anantprsd5 Kindly see this...Thank You!

sridharjajoo:
@isuruAb I had included the above code in another P.R as it was just a single line change . Hence I am closing this.:)

PR #1588, opened 2018-01-04
Images are now shown properly in landscape and portrait mode

### LINKED ISSUE #1144: Image Not Viewing Properly while toggling between Landscape and Portrait

batbrain7:
Fix #1144 

Changes: 

Screenshots for the change: 
screenshot_20171228-171826
screenshot_20171228-171850


batbrain7:
@saarthakchats I had squashed earlier commits, but I think the difference is huge therefore 4 commits are suffice to show it

gradle.properties
```diff
@@ -1,27 +0,0 @@
-# Project-wide Gradle settings.
```
abishekvashok: why was this deleted?

batbrain7: I think when I added added new changes to git I added some unnecessary files too, so while resetting the added changes I think it got deleted.
Fixing this right away

abishekvashok:
@anantprsd5 can you take a look pls

mohitmanuja:
@batbrain7 Please push the actual changes there is a lot of unnecessary changes. Please push the changes which can be clearly visible.

batbrain7:
@mohitmanuja these are the actual changes

PR #1603, opened 2018-01-06
Fixed Image Orientation problem

### LINKED ISSUE #1144: Image Not Viewing Properly while toggling between Landscape and Portrait

batbrain7:
Fix #1144 

Changes: 

Screenshots for the change: 
screenshot_20171228-171826
screenshot_20171228-171850


raj10071997:
@batbrain7- I am only giving a suggestion - why are you using this library - https://github.com/davemorrissey/subsampling-scale-image-view . Just handle the orientation changes as in the issue #1629 . Adding another library makes difficult to understand the code plus the developer has to understand the use of that library too.

batbrain7:
@mohitmanuja @anantprsd5 @isuruAb please review

anantprsd5:
@mohitmanuja Can you please take a look at this?

batbrain7:
@raj10071997 if you carefully notice using the previous photoview package there is a logcat error as : 
"org.fossasia.phimpme I/PhotoViewAttacher: ImageView no longer exists. You should not use this PhotoViewAttacher any more. "  This error was described in the issue #1144.
Due to this there were a lot more leaks during rotation. Also I did not add another library I replaced the contents of the previous photoview package with a new and updated photoview package, which is only better than the previous one and is not a redundant addition in this context, the addition is only better.

raj10071997:
@batbrain7  I only referred to the android developers documentation about handling orientation changes. The documentation says that if you don't handle orientation by yourself the system does it for you by destroying and restarting your activity. But since in manifest file there was already an "orientation" tag for SingleMediaActivity and the onConfigurationChanged() method was already defined , so it was clear we have to handle the changes by ourselves.  Read the 4th paragraph in "Handling the Configuration Change Yourself" heading below the example. Here the link - https://developer.android.com/guide/topics/resources/runtime-changes.html . It clearly says that as the configuration changes the onConfigurartionChanged() method is called and by reading fields in the Configuration (object passed as parameter in the method), you can determine the new configuration and _make appropriate changes by updating the resources used in your interface_. So we needed to just update the resources. Also If you first rotate the phone in landscape mode then open the image in SingleMediaActivity , we can't see this issue which seemed to be a clear sign that we have to handle the configuration changes. That's why I was suggesting you that we have to only handle the configuration changes.

batbrain7:
@raj10071997 I completely agree with you but also address the problem which is in the logcat : 
"org.fossasia.phimpme I/PhotoViewAttacher: ImageView no longer exists. You should not use this PhotoViewAttacher any more."
which was pointed out by @codedsun.

raj10071997:
@batbrain7  Actually if you see the file PhotoViewAttacher.java it's a log message and now read the "Remember" note "Handling the Configuration Change Yourself" heading it says - If you declare your activity to handle the orientation change and have images that should change between landscape and portrait, you must re-assign each resource to each element during onConfigurationChanged(). So when the configuration changes the imageView is destroyed by the system and that's why the log messages gets displayed in the logcat.

batbrain7:
@anantprsd5  @isuruAb the changes in the PR are done as suggested, please review and merge this

anantprsd5:
@batbrain7 The similar work was done before in another PR by @raj10071997 and I guess it has been merged and the issue is fixed. Next time onwards, I would suggest everyone to collaborate using the official Gitter channel if you are working on the same issue instead of sending different PR's. Thanks.

PR #1612, opened 2018-01-07
Fix#1601: Pinned icon in dark mode

### LINKED ISSUE #1601: Pinned to top icon not visible
**Actual Behaviour**
The pinned to top icon is not visible in dark theme.

**Expected Behaviour**
It should be visible.

Yes!

sridharjajoo:
Fix #1601 
Changes:The pinned to top icon is now visible in the dark mode as well.

Screenshots for the change: 
device 2018 01 07 130659


sridharjajoo:
@mayank8318 done the changes!

PR #1654, opened 2018-01-13
Fix #1626 :Added exit reveal animation

### LINKED ISSUE #1626: Circular reveal on toolbar
**Actual Behaviour**
Currently when a particular item is chosen by long pressing it , a circular reveal effect is seen on the toolbar from the centre to the edges . But when back button or home button is pressed there is no reveal effect from the edges to the centre which is a type of inconsistency.

**Expected Behaviour**
There should be a reveal effect from edges to the centre.

**Steps to reproduce it**
Long press any album.

Yes

sridharjajoo:
Fix #1626 

Changes: Now when the user presses the back button after selecting multiple albums or photos , a circular reveal is shown on the toolbar from outwards to inwards.


codedsun:
@isuruAb Merge this!

PR #1662, opened 2018-01-15
Fixed  #1617 - removed the blink of white color

### LINKED ISSUE #1617: blink of white color around the image when the image is zoomed in.
**Actual Behaviour**

When you zoom the image and view it for few seconds then you can see a blink of white color around the image.

**Expected Behaviour**

There should be no blink of white color when the image is zoomed in.

Yes

raj10071997:
Fixed #1617

Changes: Every time when the user touches the screen the OnUserInteraction and  OnSingleTap methods were called in SingleMediaActivity . When the SystemUI hides and fullscreen mode becomes true, the OnUserInteraction method calls startHandler method which places an unnecessary runnable in the message queue. So removed that runnable from the queue.


codedsun:
@raj10071997 Kindly update the branch 👍

Issue #1659, opened 2018-01-14
Modifications required in the search functionality.


sauravvishal8797:
**Actual Behaviour**

Issues in the search functionality.

- [ ] Soft-keyboard should automatically pop up when the focus is on search view. 

- [ ] Appropriate message should be displayed if search result is null. 

**Expected Behaviour**

Soft-keyboard should automatically pop up when the search view gets focus.
Proper message should be displayed if the search results are null.

**Steps to reproduce it**

Perform the search operation.

**Would you like to work on the issue?**

Yes.

codedsun:
@sauravvishal8797 With respect to first, we have keyboard popping up on focus to search bar. Confirmed @nikit19 .

And do we require a sort on the search? your views @mayank8318 @nikit19 Because there are many features already being present in the Phimpme and adding more features will only make the app more lagging.

mayank8318:
At this stage I don't think we need sorting or speech for searching. As these are just album names, in my opinion speech is not required here and sorting too since the no of results are bound to be few.

nikit19:
I agree with @mayank8318 and I think that the third point
 "The bottom bar should be hidden in the search mode" is also not required as the bottom bar will always be hidden under the soft keyboard

PR #1668, opened 2018-01-15
First step for #1659. Soft keyboard pops up automatically when the se…

sauravvishal8797:
…archview has focus.

First step for #1659 

Changes: Soft keyboard pops up automatically when the searchview has focus.


nikit19:
@sauravvishal8797 can you explain what you are doing here  because when I select search,the soft keyboard already pops up automatically

sauravvishal8797:
@nikit19 The soft keyboard doesn't pops up automatically for me here, so I've used the setOnQueryTextFocusChangeListener of the android searchview widget where I look for the cases when the focus is on the searchview and for the suitable cases, in the runnable object's run method the softkeyboard will keep popping up at a time delay of 200ms(which is almost negligible) so till the time the focus is on the searchview the softkeyboard will be in the popped up condition.

PR #1705, opened 2018-01-19
Fixed  #1700 on pressing back gallery is shown

### LINKED ISSUE #1700: on pressing back on camera activity, settings activity is shown.
**Actual Behaviour**

After selecting "Reset Settings" option from settings, if you select "Camera" option from the general section in settings , a dialog box appears and ask for opening the camera to support all options. Now if you select OK option from the dialog, open the camera and then press back button ,you will see that you are taken back to the same dialog box.

**Expected Behaviour**

On pressing back , Gallery should be opened.

Yes

raj10071997:
Fixed #1700

Changes: When you install the activity for the first time, open the settings and click on Camera option you will be directed to camera activity. Now if press back you will taken back to the dialog box i.e the settings activity. So added finish() to destroy the settings activity so that Gallery opens on pressing back. Same event happens when you reset your settings and click on the Camera option in settings.


raj10071997:
@codedsun  @anantprsd5  @isuruAb  @mohitmanuja Please review this PR.

PR #1707, opened 2018-01-20
Added #1665 - Favourite Icon

### LINKED ISSUE #1665: No star icon to directly add image to favourites.
**Actual Behaviour**

There is no star icon to directly add the image to favourites. Rather than this, there can be a star icon on the toolbar or the bottom bar.

**Expected Behaviour**

There should be a star icon to directly add the images to favourites.

Yes

raj10071997:
Fixed  #1665

Changes: Removed "Add to Favourite" option from the overflow menu in SingleMediaActivity and added a favourite icon on the bottom bar menu. The "Add to favourite" option is still present in the overflow menu in LFMainActivity.

Advantages: The user can now directly click on the icon to add the image to favourite and don't have to click on the overflow menu to add it to favourites. Another factor, generally in apps if we have the option to add something to favourite, then generally a star or heart icon is present there.

Gif for the change: 

fav


raj10071997:
@codedsun  @mohitmanuja  @anantprsd5  @isuruAb Please review this PR.

PR #1757, opened 2018-01-27
Fix #1750-All the Snackbar messages are displayed at the bottom.

### LINKED ISSUE #1750: Snackbar display issue.
**Actual Behaviour**

For some cases the snackbar messages are displayed at the bottom of the screen covering the bottom bar, while for some cases those are displayed above the bottom bar.

**Expected Behaviour**

All the snackbar messages should be displayed at the bottom of the screen,  looks better in that way.

**Steps to reproduce it**

Perform operations like zip, add to favourites, copy, move etc.

Yes.

sauravvishal8797:
Fixed #1750 

Changes: All the Snackbar messages are displayed at the bottom of the screen.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
@@ -542,14 +542,14 @@ public class SingleMediaActivity extends SharedMediaActivity implements ImageAda
                             handleEditorImage(data);
                             if (ContentHelper.copyFile(getApplicationContext(), new File(imageUri.getPath()), new File(getAlbum().getPath()))) {
                                 //((ImageFragment) adapter.getRegisteredFragment(getAlbum().getCurrentMediaIndex())).displayMedia(true);
-                                SnackBarHandler.showWithBottomMargin(parentView, getString(R.string.new_file_created), bottomBar.getHeight());
+                                SnackBarHandler.show(parentView, getString(R.string.new_file_created));
                             }
                             //adapter.notifyDataSetChanged();
                         } catch (Exception e) {
                             Log.e("ERROS - uCrop", imageUri.toString(), e);
                         }
                     } else
-                        SnackBarHandler.showWithBottomMargin(parentView, "errori random", bottomBar.getHeight());
+                        SnackBarHandler.show(parentView, "errori random");
                     break;
                 default:
                     break;
@@ -696,7 +696,7 @@ public class SingleMediaActivity extends SharedMediaActivity implements ImageAda
                         else{
                             getAlbum().copyPhoto(getApplicationContext(), getAlbum().getCurrentMedia().getPath(), path);
                             bottomSheetDialogFragment.dismiss();
-                            SnackBarHandler.showWithBottomMargin(relativeLayout, getString(R.string.copied_successfully) + " to " + path, bottomBar.getHeight());
+                            SnackBarHandler.show(relativeLayout, getString(R.string.copied_successfully));
                         }
                     }
                 });
@@ -727,7 +727,7 @@ public class SingleMediaActivity extends SharedMediaActivity implements ImageAda
                     editIntent.putExtra("requestCode", ACTION_REQUEST_EDITIMAGE);
                     startActivity(editIntent);
                 } else
-                    SnackBarHandler.showWithBottomMargin(parentView, getString(R.string.image_invalid), bottomBar.getHeight());
+                    SnackBarHandler.show(parentView, getString(R.string.image_invalid));
                 break;
 
             case R.id.action_use_as:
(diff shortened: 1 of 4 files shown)


codedsun:
@sauravvishal8797 Kindly update the branch .

app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
```diff
@@ -542,14 +542,14 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                             handleEditorImage(data);
                             if (ContentHelper.copyFile(getApplicationContext(), new File(imageUri.getPath()), new File(getAlbum().getPath()))) {
                                 //((ImageFragment) adapter.getRegisteredFragment(getAlbum().getCurrentMediaIndex())).displayMedia(true);
-                                SnackBarHandler.showWithBottomMargin(parentView, getString(R.string.new_file_created), bottomBar.getHeight());
```
codedsun: What was the error in this method, are we getting the 
`bottomBar.getHeight()` some  positive value?. Please check it out !

nikit19:
In my opinion all snackBar messages should be displayed with the bottom margin so that it doesn't overlap the bottom bar.
@sauravvishal8797 If there is any snackBar which is displayed at the bottom of the screen it should also be displayed above the bottom bar
@mayank8318 If I am right this is the very reason why this helper method was made ?

nikit19:
https://github.com/fossasia/phimpme-android/pull/1584
I remember sending a PR to fix this issue because snackBars weren't visible without the helper method.
Why are you reverting it back? Moreover the snackBars are not visible in singleMedia Activity,I tried running your PR locally

sauravvishal8797:
@nikit19 Some of the snackbar messages were displayed at the bottom covering the bottom bar, while others were displayed above the bottom bar, so I guess all the snackbar messages should be displayed in same manner, so we could either show it covering the bottom bar i.e at the bottom of the screen or we could show it above the bar. And showing it at the bottom looked better to me and moreover these messages are displayed for a really short interval of time so I think it shouldn't be much of a problem if it covers the bottom bar for such interval of time. What do you think? And can you mention which of the snackbar messages are not visible in the SingleMediaActivity, because for me its working fine.

nikit19:
@sauravvishal8797 You are right,all snackbar messages should be displayed in a similar manner.
But when you display it at the bottom ,the bottom bar does not get fully covered and it looks weird.
I guess that is why this helper method was made. So if there are any snackBar messages that are displayed at the bottom they should be displayed above the bottom bar.

What Android version are you using? I don't see the snackBar messages in the SingleMedia activity, can you please send a screenshot of how they are displayed in the singleMedia Activity

SnackBar messages are either not visible at all or seen this like this when I try  your PR

screenshot_20180127-225634

sauravvishal8797:
@nikit19 I'm using Android version 6.0.1, its displayed fine for me as shown below. What version are you running on?
screenshot_2018-01-27-23-44-15-663_org fossasia phimpme

nikit19:
@sauravvishal8797  I use 7.0
May be @codedsun or @mayank8318 could confirm if it is working 
But even without these changes ,the snackBar is displayed like this in singleMedia activity
or is it displayed differently in your phone?

sauravvishal8797:
@nikit19 Without the current changes for some operations snackbar messages are displayed above the bottom bar while for some they are displayed at the bottom covering the bottom bar i.e some snackbars are displayed using show helper method while some are displayed with showwithMarginBottom helper method which should not be the case, every snackbar message should be displayed either using the show method or the showwithmarginBottom method.
screenshot_2018-01-28-00-55-05-435_org fossasia phimpme

So what I wanted to do is

nikit19:
This looks good. In my opinion, snackBar messages should be displayed using showWithMarginBottom method. So that all views can be displayed without overlapping the bottom bar

sauravvishal8797:
Ok. I guess if even for a single device displaying at the bottom does not works, then we should probably go with showwithmarginBottom method. Will update the PR to use showwithMarginBottom method everywhere.

codedsun:
As @nikit19 stated, the helper method were made as previously the Snackbar was colliding with the bottom bar and hence the Snackbar was not displayed properly. Hence, according to my opinion, why they should be referenced using the same methods (like showWithMarginBottom( ) or make( ) )until we are finding some issues.

@mayank8318 Ur opinion!?

mayank8318:
In my device the snackbars are not visible if you dont use the bottomMargin helper method and so I made this method. Also instead of covering the bottom navigation bar, i think showing the Snackbar over it is a better idea since it is clearly visible and very intuitive.

sauravvishal8797:
@nikit19 @codedsun @mayank8318 Ok. Then we should go with the bottomMargin method, but for some operations like zip, adding to favourites etc the snackbar is displayed at the bottom covering the bottom bar, so I guess for those cases also we should change it to display the snackbar above the bottom bar.

mayank8318:
In my view that would be the best thing to do.
@nikit19 @codedsun @mohitmanuja

nikit19:
Yeah agreed bottomMargin helper method should be used everywhere

abishekvashok:
Closing due to inactiviry

PR #1769, opened 2018-01-29
Fix #1767-App crash on launch fixed.

### LINKED ISSUE #1767: App crashes on launch.
**Actual Behaviour**

App crashes on launch.

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

Open the application .

**LogCat for the issue**

FATAL EXCEPTION: main
                                                                      Process: org.fossasia.phimpme, PID: 12961
                                                                      java.lang.NullPointerException: Attempt to invoke interface method 'android.view.MenuItem android.view.MenuItem.setVisible(boolean)' on a null object reference
                                                                          at org.fossasia.phimpme.gallery.activities.LFMainActivity.onPrepareOptionsMenu(LFMainActivity.java:1586)
                                                                          at android.app.Activity.onPreparePanel(Activity.java:2919)
                                                                          at android.support.v4.app.FragmentActivity.onPrepareOptionsPanel(FragmentActivity.java:535)
                                                                          at android.support.v4.app.FragmentActivity.onPreparePanel(FragmentActivity.java:523)
                                                                          at android.support.v7.view.WindowCallbackWrapper.onPreparePanel(WindowCallbackWrapper.java:103)
                                                                          at android.support.v7.app.AppCompatDelegateImplBase$AppCompatWindowCallbackBase.onPreparePanel(AppCompatDelegateImplBase.java:362)
                                                                          at android.support.v7.view.WindowCallbackWrapper.onPreparePanel(WindowCallbackWrapper.java:103)
                                                                          at android.support.v7.app.ToolbarActionBar$ToolbarCallbackWrapper.onPreparePanel(ToolbarActionBar.java:567)
                                                                          at android.support.v7.app.ToolbarActionBar.populateOptionsMenu(ToolbarActionBar.java:455)
                                                                          at android.support.v7.app.ToolbarActionBar$1.run(ToolbarActionBar.java:61)
                                                                          at android.os.Handler.handleCallback(Handler.java:739)
                                                                          at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                          at android.os.Looper.loop(Looper.java:148)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5441)
                                                                          at java.lang.reflect.Method.invoke(Native Method)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:738)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:628)

Yes.

sauravvishal8797:
Fixed #1767 

Changes: Print photo option was also visible in the LFMainActivity which was causing the crash. Removed print option from LFMainActivity.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/util/SecurityHelper.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/util/SecurityHelper.java
@@ -51,7 +51,7 @@ public class SecurityHelper {
         String secur = SP.getString("SecuredLocalFolders", "");
         Gson gson = new Gson();
         String[] secfo = gson.fromJson(secur, String[].class);
-        if(secfo.length!=0){
+        if(secfo!=null){
             this.securedfolders = secfo;
         }
         this.activeSecurity = SP.getBoolean(context.getString(R.string.preference_use_password), false);
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -1583,7 +1583,6 @@ public class LFMainActivity extends SharedMediaActivity {
         menu.findItem(R.id.action_copy).setVisible(visible);
         menu.findItem(R.id.action_move).setVisible((visible || editMode)&&!fav_photos);
         menu.findItem(R.id.action_add_favourites).setVisible((visible || editMode)&&(!albumsMode&&!fav_photos));
-        menu.findItem(R.id.print).setVisible(visible);
         menu.findItem(R.id.excludeAlbumButton).setVisible(editMode && !all_photos && albumsMode && !fav_photos);
         menu.findItem(R.id.zipAlbumButton).setVisible(editMode && !all_photos&&albumsMode &&!fav_photos && !hidden &&
                 getAlbums().getSelectedCount() == 1);
@@ -1641,19 +1640,6 @@ public class LFMainActivity extends SharedMediaActivity {
                     }});
                 detailsDialog.show();
                 AlertDialogsHelper.setButtonTextColor(new int[]{DialogInterface.BUTTON_POSITIVE}, getAccentColor(), detailsDialog);
-                return true;
-
-            case R.id.print:
-                if(!all_photos){
-                    for(int i = 0; i < getAlbum().getSelectedMedia().size(); i++){
-                        PrintHelper photoPrinter = new PrintHelper(this);
-                        photoPrinter.setScaleMode(PrintHelper.SCALE_MODE_FIT);
-                        Bitmap bitmap = BitmapFactory.decodeFile(getAlbum().getSelectedMedia(i).getPath(), new
-                                BitmapFactory.Options
-                                ());
-                        photoPrinter.printBitmap(getString(R.string.print), bitmap);
-                    }
-                }
                 return true;
 
             case R.id.select_all:


raj10071997:
@sauravvishal8797 the app is still crashing even after making the changes. In logcat, the error is this -


01-30 07:46:08.987 2543-2543/org.fossasia.phimpme E/AndroidRuntime: FATAL EXCEPTION: main
                                                                    Process: org.fossasia.phimpme, PID: 2543
                                                                    java.lang.RuntimeException: Unable to start activity ComponentInfo{org.fossasia.phimpme/org.fossasia.phimpme.gallery.activities.LFMainActivity}: **_java.lang.NullPointerException: Attempt to get length of null array
                                                                        at_** android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2756)
                                                                        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2825)
                                                                        at android.app.ActivityThread.-wrap12(ActivityThread.java)
                                                                        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1557)
                                                                        at android.os.Handler.dispatchMessage(Handler.java:110)
                                                                        at android.os.Looper.loop(Looper.java:203)
                                                                        at android.app.ActivityThread.main(ActivityThread.java:6343)
                                                                        at java.lang.reflect.Method.invoke(Native Method)
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1084)
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:945)
                                                                     Caused by: java.lang.NullPointerException: Attempt to get length of null array
                                                                        at org.fossasia.phimpme.gallery.util.SecurityHelper.updateSecuritySetting(SecurityHelper.java:54)
                                                                        at org.fossasia.phimpme.gallery.util.SecurityHelper.<init>(SecurityHelper.java:37)
                                                                        at org.fossasia.phimpme.gallery.activities.LFMainActivity.onCreate(LFMainActivity.java:663)
                                                                        at android.app.Activity.performCreate(Activity.java:6689)
                                                                        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1140)
                                                                        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2709)




So, I made the changes in SecurityHelper.java to ensure that the array is not null, then check the length of the array and then the app was starting. Please see if this is also the case with you.

raj10071997:
@sauravvishal8797  Can you please uninstall the app and then install the app with changes as you mentioned. It is Giving null error for password settings. Do you have password set for your app?

sauravvishal8797:
@raj10071997 Yea! After uninstalling and re-installing the app crashes. As the updatesecuritysetting method in the SecurityHelper class is called everytime on start of LFmainactivity, we need to check for the secured folders list to be null to avoid the app crash on the install. Will update the PR to check for the null case, and thanks for pointing it out. 👍

app/src/main/java/org/fossasia/phimpme/gallery/util/SecurityHelper.java
```diff
@@ -51,7 +51,7 @@ public void updateSecuritySetting(){
         String secur = SP.getString("SecuredLocalFolders", "");
         Gson gson = new Gson();
         String[] secfo = gson.fromJson(secur, String[].class);
-        if(secfo.length!=0){
+        if(secfo!=null){
```
raj10071997: this condition only checks whether the string is not null but it doesn't check if it is empty, which means the body of the if condition can be executed even if the secfo string is empty. You can first check that the string is not null to avoid the crashing of the app and then check if the string is empty or not.

sauravvishal8797: secfo list will only have non-null value when securing local folders option is enabled, so there will one or more folders to secure thereby the secfo list will not be empty in that case so I guess only checking for secfo to be null will do.

raj10071997: But what if you reset the settings, then all the values in the shared preference will be removed and it will return an empty string for secur which was in the case when you install the app.

sauravvishal8797: On resetting the settings, the secur string will store an empty string, which when passed to the gson.fromJson() function will return a null object, so it won't execute the statement in the if block and as the securedfolders array is only required when the security for local folders are enabled and as on resetting, the security is disabled so again just checking for null will do. Whenever the security for the local folders is enabled the secur string wouldn't  be empty and therefore passing it to the fromJson() function will return an array with path of the folders which are to be secured.

sauravvishal8797:
@anantprsd5 @mohitmanuja @isuruAb Please merge this commit, this provides solution for the app crashing on the launch itself, which is required to be resolved as soon as possible for further development of the application.

PR #1808, opened 2018-02-05
Fixed  #1749 MaterialShowCase is now showing up immediately

### LINKED ISSUE #1749: MaterialShowCase which guides the user about the gallery, account and camera buttons is little slow
**Actual Behaviour**

When you install the app, MaterialShowCase that guides the user about the below navigation bar is little slow. As soon as you install and open the app press any button say settings in the overflow menu , you will see that the dialogs starts appearing in the settings or when you come back from settings.

**Expected Behaviour**

The dialog should start before the user can take any action of if he/she takes then the user should have the option to cancel it.

Yes.

raj10071997:
Fixed #1749

Changes: Actually before the delay was 4 seconds after layout of LFMainActivity.java was visible, during which any action could be taken by the user. Now the delay is 500 milliseconds which doesn't allow the user to take any action before the MaterialShowCase starts.


raj10071997:
@isuruAb @abishekvashok You can merge this branch I have updated it.

app/src/main/java/org/fossasia/phimpme/base/BaseActivity.java
```diff
@@ -77,7 +77,7 @@ protected void onResume() {
     private void presentShowcaseSequence() {
 
         ShowcaseConfig config = new ShowcaseConfig();
-        config.setDelay(4000); // half second between each showcase view
+        config.setDelay(500); // half second between each showcase view
```
abishekvashok: Looking at the comment I can see a delay of half second. But this is 5 seconds. 50 should be used here right?

raj10071997: @abishekvashok No the setDelay() method takes long integer in form of milliseconds. 500 is for 0.5 milliseconds.

abishekvashok: Sorry I forgot that 1000 milliseconds formed a second!

raj10071997: it's Ok.

PR #1830, opened 2018-02-09
Fix #1820-Error message displayed when entering wrong password.

### LINKED ISSUE #1820: Error message for entering wrong password not displayed for some cases.
**Actual Behaviour**

After enabling the security for delete, local folders, hidden folders if the users enters wrong password no message is displayed as the floating label.

**Expected Behaviour**

Floating error message should be displayed when the user enters the wrong password just like it is displayed for the security option in the settings page. 

**Steps to reproduce it**

Enable security for delete option, local folders etc, and try entering the wrong password.

Yes.

sauravvishal8797:
Fixed #1820 
Changes: Error message is displayed as a floating label in the edittexts when entering wrong password, the floating label(error message) dissappears when the user starts typing the password once again after the failed attempt.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -463,7 +463,8 @@ public class LFMainActivity extends SharedMediaActivity {
         @Override
         public boolean onLongClick(View v) {
             final Album album = (Album) v.findViewById(R.id.album_name).getTag();
-            if (securityObj.isActiveSecurity() && securityObj.isPasswordOnfolder()) {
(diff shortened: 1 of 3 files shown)


sauravvishal8797:
@mayank8318 I guess TextInputLayout is already there for the edittexts, and if you're talking about the listeners, I've used the edittext's addTextChangeLIstener interface to make sure that the wrong password message is not visible while typing the password, it is only displayed if the user tries to proceed after typing in the wrong password. Earlier once the user entered the wrong password, the wrong password message is displayed but when the user starts to re-enter the password the wrong password message is still visible which should not be case.

sauravvishal8797:
@abishekvashok Resolved the conflicts, please merge if no changes required.

abishekvashok:
Please wait for Travis!

PR #1864, opened 2018-02-18
Fix#1862:LocalFolder Checkbox Fix

### LINKED ISSUE #1862: Security Activity issue
**Actual Behaviour**

The local folder toggle button is set to true, even when no folder is selected and the app crashes on opening the folders

**Expected Behaviour**

It shouldn't change and the app shouldn't crash.

**Steps to reproduce it**

Go to the Settings, follow the security use any password setup.
And then click on local folders, then don't select any folder nd press back the toggle button is still set to true.

Yes!

codedsun:
Fixed #1862

Changes: 
1.Refractored the layout
2.Added on back pressed listener whose absence was creating the app crash.


codedsun:
@isuruAb Check and merge!

abishekvashok:
@codedsun pls resolve conflicts.

app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
```diff
@@ -131,8 +131,8 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
                     title.setText("Choose folders to secure");
                     RecyclerView recyclerView = (RecyclerView) view.findViewById(R.id.secure_folder_recyclerview);
                     recyclerView.setLayoutManager(new LinearLayoutManager(getApplicationContext()));
-                    SecureDialogAdapter s = new SecureDialogAdapter();
-                    recyclerView.setAdapter(s);
+                    SecureDialogAdapter securedLocalFolders = new SecureDialogAdapter();
```
abishekvashok: If possible make this final.

raj10071997: I think there is no need to make it final as the adapter is not changing after the initialization.

abishekvashok: It speeds up execution if you do so.
https://developer.android.com/training/articles/perf-tips.html

raj10071997: Yes, you are right, declaring final does improve performance. But in local variable which is declared and accessed only once, that difference in performance will be negligible in this case.

abishekvashok: 1 maybe negligible compared to 1000. But 1+1+1+1+1+1... (100 times may not). It was just a suggestion though. I am okay to merge this at all costs :)

raj10071997: Well it was a nice suggestion even I learned something.

abishekvashok: Ping @codedsun

codedsun:
@abishekvashok Merge this!

codedsun:
@abishekvashok Please approve the PR!

PR #1879, opened 2018-02-21
Added the Gamma Effect

### LINKED ISSUE #1878: Add Gamma Effect
**New Feature**

Add a new filter in the editor of the app that shows the gamma effect.

Yes.

batbrain7:
Fixed #1878 

Changes: Added the Gamma Effect into the app.

Screenshots for the change: 

Before :
 
screenshot_20180221-191752

After : 

screenshot_20180221-191843


app/src/main/jni/enhance.h
```diff
@@ -1,6 +1,9 @@
 
 #include <opencv2/core/core.hpp>
 #include <opencv2/imgproc/imgproc.hpp>
+#include <stdlib.h>
```
abishekvashok: Why do you need these libs?

batbrain7: @abishekvashok I was trying a different method before which required stdlib.h

abishekvashok: Hm

Issue #1634, opened 2018-01-10
Issue in Image details view.


sauravvishal8797:
**Actual Behaviour**

The image titles exceeding single line are not visible properly.

**Expected Behaviour**

The image titles exceeding single line should be visible properly.

**Steps to reproduce it**

Open details for any image.

**Screenshots of the issue**

![screenshot_2018-01-10-19-17-00-626_org fossasia phimpme](https://user-images.githubusercontent.com/20841578/34781847-f12addb6-f64c-11e7-806a-4e11df089ec7.png)

**Would you like to work on the issue?**

Yes.

Shubham2320:
@sauravvishal8797 can I work on it?

sauravvishal8797:
@Shubham2320 I have already referenced a PR with this issue. The issue is fixed.

PR #1881, opened 2018-02-22
Fix #1850 - information is now properly with their tags.

### LINKED ISSUE #1850: Infomation with their tags is not assigned properly in details of an image.
**Actual Behaviour**

See the below screenshot for the issue. The title of the tag title is not aligned properly. Similarly, Path is also not aligned properly with it's tag.

**Expected Behaviour**

All the information should be aligned properly with their tags.

**Screenshots of the issue**

details

Yes.

raj10071997:
Fixed #1850

Changes: The information is now properly assigned with their respective tags. Also added a "2dp" bottom margin to avoid the problems faced in the issue #1634. Since I added a "2dp" bottom margin so reduce the height of the view (from 10dp to 8dp) that provides space between tags in details view. Also assigned a common textSize to the information in details view.

Screenshots for the change: 
detailschanges


raj10071997:
@abishekvashok  Thanks :smile:  . @sauravvishal8797  Can you please confirm these changes as you were facing the issue #1634 ?

PR #1883, opened 2018-02-22
Fix #1822 Removed settings option from the overflow menu

### LINKED ISSUE #1822: Two settings options in LFMainActivity
**Actual Behaviour**
There are two ways to open settings in LFMainActivity. One is by overflow menu and the other is by navigation drawer, this creates a redundancy.

**Expected Behaviour**

There should be only one way to open settings.

Yes.

raj10071997:
Fixed #1822

Changes: Now there is only one way to open settings, that is, from the navigation drawer. Removed this option from the overflow menu when you are in AlbumsMode.
Screenshots for the change: 

settings

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -1648,6 +1648,7 @@ public class LFMainActivity extends SharedMediaActivity {
             menu.setGroupVisible(R.id.photos_option_men, false);
             menu.findItem(R.id.all_photos).setVisible(!editMode && !hidden);
             menu.findItem(R.id.search_action).setVisible(!editMode);
+            menu.findItem(R.id.settings).setVisible(false);
 
             if (getAlbums().getSelectedCount() >= 1) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -1648,6 +1648,7 @@ public boolean onPrepareOptionsMenu(final Menu menu) {
             menu.setGroupVisible(R.id.photos_option_men, false);
             menu.findItem(R.id.all_photos).setVisible(!editMode && !hidden);
             menu.findItem(R.id.search_action).setVisible(!editMode);
+            menu.findItem(R.id.settings).setVisible(false);
```
abishekvashok: Can you remove it from the menu instead of setting it hidden? Or are there any other trouble?

raj10071997: First I thought of the same thing of removing it. But If I remove it, the settings option will not be available in the overflow menu when you open an album.

abishekvashok: Thanks!

PR #1887, opened 2018-02-22
Fix #1886-Issue in photos long press operation resolved.

### LINKED ISSUE #1886: Issue in photos long select operation.
**Actual Behaviour**

Settings option is visible in the overflow menu after performing long press on photos i.e in the edit mode.

**Expected Behaviour**

No need to show settings option in the menu in edit mode, creates redundancy. 

**Steps to reproduce it**

Open any album, perform long press operation then go to the overflow menu.

Yes.

sauravvishal8797:
Fixed #1886 

Changes: Settings option is not visible in the overflow menu on performing long press operation on photos.


abishekvashok:
Thanks!

PR #1897, opened 2018-02-25
Image share from Upload History Fix

### LINKED ISSUE #1461: Option to Share Image again from upload history
**Actual Behaviour**

Their is no option to delete upload history

**Expected Behaviour**

- [ ] Has option to again share the same image.

 

Yes!

codedsun:
Fixed #1461 : Share an image again from Upload History

Changes:Added and onClickListener()

Screenshots for the change:


sauravvishal8797:
@codedsun The imageshare icon displays sort of the platform through which the image is shared(like whatsapp, fb etc icons will be displayed when performing the share operation through the SingleMediaActivity), so instead of enabling the share feature via clicking on the imageshare icon, you could do it via the long press operation on photos i.e on performing the long press operation on photos , the photos will be displayed in the edit-mode and share option could be displayed on the toolbar. Through this method sharing multiple photos from upload history at once can also be implemented.

codedsun:
@sauravvishal8797  I though of that but we are having a case, where the user deletes the shared image, still we are having the path in the realm of that previous pics and when I thought of removing that, it's of no use as it gives an extra info to the user to what he shared and even if he deletes the shared pic from the original path he could find in the upload history. Hence if we try to implement it, it will make the issue more complex.

codedsun:
@sauravvishal8797 Updated!

PR #1900, opened 2018-02-25
Added option to unfavourite image

### LINKED ISSUE #1884: Add option to unfavourite images
**Actual Behaviour**
Currently there is no option to unfavourite any image. An option can be added to the overflow menu when a favourite image is selected if the user wants to unfavourite it or not.

Yes!

sridharjajoo:
Fixed #1884 

Changes: The user can unfavourite the images by long pressing on them and clicking the unfavourite option from the overflow menu.

Screenshots for the change: 
un


sridharjajoo:
@abishekvashok Kidly review!

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -1686,13 +1690,14 @@ public boolean onPrepareOptionsMenu(final Menu menu) {
...
         menu.findItem(R.id.action_copy).setVisible(visible);
         menu.findItem(R.id.action_move).setVisible((visible || editMode) && !fav_photos);
+        menu.findItem(R.id.unfavourite_image).setVisible((!albumsMode) && (editMode || visible));
         menu.findItem(R.id.action_add_favourites).setVisible((visible || editMode) && (!albumsMode && !fav_photos));
         menu.findItem(R.id.excludeAlbumButton).setVisible(editMode && !all_photos && albumsMode && !fav_photos);
         menu.findItem(R.id.zipAlbumButton).setVisible(editMode && !all_photos && albumsMode && !fav_photos && !hidden &&
                 getAlbums().getSelectedCount() == 1);
         menu.findItem(R.id.select_all).setVisible(editMode);
         menu.findItem(R.id.delete_action).setVisible((!albumsMode || editMode) && (!all_photos || editMode) &&
-                (!fav_photos || editMode));
```
abishekvashok: Why was this removed?

sridharjajoo: The delete icon was present in the favourites mode and its function was to unfavourite an image. But as its was a delete icon it was a bit confusing for the user what action will it perform . So i remove this line so that the toolbar will not have the delete icon and added the option to unfavourite image from overflow menu.

abishekvashok: Maybe you could change it to an unfavourite button instead of a delete. But it would be used in other areas as well I guess.

sridharjajoo: So should I add the option to unfavourite an image to the toolbar or keep it in the overflow menu only?

abishekvashok: Okay then can you remove this from the menu then?

abishekvashok: Nope. We will go with the same way then.

yashLadha:
I didn't get the idea of the image that is not favourite ans showing the option of unfavourite in the menu. Instead of inflating a static menu, it should be according to the type of images the user has selected. @sridharjajoo

sridharjajoo:
When an individual image is selected which is not added to favourites , then the option to unfavourite it will not apper , only the option to favourite it will appear in the overflow menu. But when a group of images containing both already favourited images as well as non favourited images are selected then both the option to favourite and unfavourite it will appear.

yashLadha:
But still the idea is to show only when a number of favorite images are selected and not including any unfavorite image.

sridharjajoo:
@abishekvashok kindly merge this approved pr..thanks!

abishekvashok:
It would be much better the way @yashLadha suggested. But I think we can do this now and do it in another pr.

PR #1904, opened 2018-02-26
Fix #1882 - fixed details view in landscape mode

### LINKED ISSUE #1882: app crashes on opening details view.
**Actual Behaviour**

When opening the details of an image in landscape mode the app crashes. 

**Expected Behaviour**

The app shouldn't crash.

Yes.

raj10071997:
Fixed #1882

Changes: The details view couldn't be opened in landscape mode because the layout for landscape mode didn't contain the ids' for some view, like the view for details icon didn't contain the id- "@+id/detail_icon", location icon also didn't contain it's id.
Some other views also didn't contain their respective ids' which were present in the layout for portrait mode but not in landscape mode. But the problem was not just because of the ids'. When you click the details and open the details layout and as soon you turn into landscape mode, you go back to the image and then you have to again open the details view to view the details. Actually when you rotate the screen the onConfigurationChanged() method was called and in that method, the code to handle the details view in landscape mode was not written due to which the Activity starts again and we are directed back to the image from the details view, when the orientation changes. So made some changes to add the ScrollView in details layout(image_description.xml) so that in portrait mode as well as in landscape mode the information can be viewed in the same layout. Also a ScrollView was needed in portrait layout of the details layout(image_description.xml) because if the description is too long the whole description couldn't be seen in this case (See the first screenshot for this issue). See the below gif for the changes.

Screenshots for the change: 
detail

Gif for the change: 
When you open details in landscape mode.
gif1

Gif for the change: 
When you open details in portrait mode.
gif2


raj10071997:
@abishekvashok You can merge this branch I have updated it.

PR #1905, opened 2018-02-26
Fix #1621 - Snackbar message displayed after successful delete operation.

### LINKED ISSUE #1621: Display Toast when photo is deleted
**Actual Behaviour**
There is no message displayed when a photo is deleted. 

**Expected Behaviour**
There should be a message displayed (eg Toast) when a photo is deleted, to indicate that the deletion was successful. 

**Steps to reproduce it**
Open a photo in the gallery and delete it. 

yes

Ananthusubramanian:
Fixed #1621

Changes: Added Snackbar after deletion of the photo


Ananthusubramanian:
Can anyone review this PR

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -1988,7 +1993,11 @@ protected void onPostExecute(Boolean result) {
                                     } else
                                         mediaAdapter.swapDataSet(getAlbum().getMedia());
                                 } else if (all_photos && !fav_photos) {
-                                    clearSelectedPhotos();
+                                    if(succ) {
```
raj10071997: Why did you remove clearSelectedPhotos() method call?

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -1999,6 +2008,11 @@ protected void onPostExecute(Boolean result) {
                                     clearSelectedPhotos();
                                     getfavouriteslist();
                                     new FavouritePhotos().execute();
+                                    if(succ) {
```
raj10071997: Instead of checking for the same condition -"if(succ)", every time in each of the if and else conditions, you can check it for once only. That will also do the work and the code will be less.

Ananthusubramanian: The function declares true to variable succ after the function ends, so even if it isn't deleted it would make "succ=true".Thats why I did like that.

raj10071997: Can you specify the line that make succ = true even the file is not deleted? Because in my view succ variable can have both values false and true. Only if we delete the file from Favourite section the succ variable is assigned the value true in background. I am just saying that for now just write once that if and else condition of yours and remove that string from strings.xml that says photo deleted from favourites as there is already an issue regarding unfavourite option.

abishekvashok: @Ananthusubramanian you are putting that block into three places. It can be one. Right? That's what @raj10071997 suggested. Can you do that? :)

Ananthusubramanian: Yeah I will do that :)

abishekvashok: Okay thanks :)

Ananthusubramanian:
@abishekvashok  I have changed it to a function ,so as to reduce to code .Does this look fine?

Ananthusubramanian:
@abishekvashok @batbrain7  Resolved the conflicts

PR #1910, opened 2018-02-28
Bug in TextStickerView resolved

### LINKED ISSUE #1902: Text going outside of the image
**Actual Behaviour**
The text which is written on the image as an edit option goes outside the image area as shown in the following figure.
textoutside

**Expected Behaviour**
The text shouldn't go outside the image area.

**Steps to reproduce it**
Go to the edit mode of the app.

sridharjajoo:
Fixed #1902 
Changes: When text is written on an image while editing it , the text doesn't go outside the given area.
device-2018-03-01-004405


app/src/main/java/org/fossasia/phimpme/editor/fragment/AddTextFragment.java
```diff
@@ -216,10 +216,14 @@ public void backToMain() {
     public void onShow() {
         activity.changeMode(EditImageActivity.MODE_TEXT);
         activity.mainImage.setImageBitmap(activity.mainBitmap);
+        activity.addTextFragment.getmTextStickerView().mainImage = activity.mainImage;
+        activity.addTextFragment.getmTextStickerView().mainBitmap = activity.mainBitmap;
         mTextStickerView.setVisibility(View.VISIBLE);
         mInputText.clearFocus();
     }
 
+    public TextStickerView getmTextStickerView(){return mTextStickerView;}
```
abishekvashok: Can this be made static?

yashLadha:
How can we select the edges that are not in the view (hidded in the black background that is out of the image) to resize the image?

sridharjajoo:
@isuruAb @abishekvashok Kindly review the P.R.

sridharjajoo:
@anantprsd5 @abishekvashok kindly review and merge

app/src/main/java/org/fossasia/phimpme/editor/fragment/AddTextFragment.java
abishekvashok: Why do u think so?

sridharjajoo: Making the function static would lead to conversion of other variables to static as well . Also the same approach (no static function) has been used in case when the paint and the text was going out of bounds as well.

isuruAb:
@abishekvashok  please review this

PR #1912, opened 2018-02-28
Fix #1659-Appropriate message is displayed if search results are null.

### LINKED ISSUE #1659: Modifications required in the search functionality.

sauravvishal8797:
Fixed #1659 
Changes: Appropriate message is displayed if the results are null on performing the search folder operation.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -216,6 +216,8 @@ public class LFMainActivity extends SharedMediaActivity {
     protected View toolbari;
     @BindView(R.id.nothing_to_show)
     protected TextView nothingToShow;
+    @BindView(R.id.no_search_results)
+    protected TextView textView;
     @BindView(R.id.Drawer_Default_Icon)
     protected IconicsImageView defaultIcon;
     @BindView(R.id.Drawer_hidden_Icon)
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -1577,6 +1582,12 @@ private void checkNothingFavourites() {
             starImageView.setColorFilter(ContextCompat.getColor(this, R.color.black), PorterDuff.Mode.SRC_ATOP);
     }
 
+    private void checkNoSearchResults(String result){
+        textView.setText(getString(R.string.null_search_result) + " " + '"' + result + '"' );
```
abishekvashok: In the string there's already space at the end, do we need two spaces?

sauravvishal8797: @abishekvashok Removed the space at the end of the string declaration in strings.xml file. Please review.

Issue #1825, opened 2018-02-09
Cluttered animation while switching Activities


yashLadha:
**Actual Behaviour**

Same entry animation is being applied on different fragments, because of which ui is looking very much cluttered.

**Expected Behaviour**

The transition between different views should be smooth.

**Steps to reproduce it**

Just switch over between Camera, Gallery and Accounts.

**LogCat for the issue**

None

**Screenshots of the issue**

![image](https://user-images.githubusercontent.com/18033231/36033892-340e8d42-0dd8-11e8-83a9-7435344d2299.png)

**Would you like to work on the issue?**

Yes

Pravesh-Jamgade:
by smooth transition ,means you want instant change or a delay screen

yashLadha:
By smooth means, i means it should not the show black screen while transition. As you can see in the gif @Pravesh-Jamgade . I am not getting time to make adjustments. Will send PR in weekend.

PR #1914, opened 2018-03-01
Smoothened transition between base activities.

yashLadha:
Fixes: #1825

Changes: Smoothened animation while switching activities.

Screenshots for the change: 

anim-opt

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
@@ -140,8 +140,9 @@ public class AccountActivity extends ThemedActivity implements AccountContract.V
         ButterKnife.bind(this);
         ActivitySwitchHelper.setContext(this);
         parentLayout.setBackgroundColor(getBackgroundColor());
-        this.overridePendingTransition(R.anim.right_to_left,
-                               R.anim.left_to_right);parentLayout.setBackgroundColor(getBackgroundColor());
+        overridePendingTransition(R.anim.right_to_left,
+                R.anim.left_to_right);
+        parentLayout.setBackgroundColor(getBackgroundColor());
         accountAdapter = new AccountAdapter();
         accountPresenter = new AccountPresenter(realm);
         phimpmeProgressBarHandler = new PhimpmeProgressBarHandler(this);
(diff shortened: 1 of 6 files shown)


yashLadha:
Done @sridharjajoo . Please Update your review.

yashLadha:
Ping @mariobehling

isuruAb:
@abishekvashok  please review

app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -140,8 +140,8 @@ public void onCreate(Bundle savedInstanceState) {
         ButterKnife.bind(this);
         ActivitySwitchHelper.setContext(this);
         parentLayout.setBackgroundColor(getBackgroundColor());
-        this.overridePendingTransition(R.anim.right_to_left,
-                               R.anim.left_to_right);parentLayout.setBackgroundColor(getBackgroundColor());
+        overridePendingTransition(R.anim.right_to_left,
+                R.anim.left_to_right);parentLayout.setBackgroundColor(getBackgroundColor());
```
abishekvashok: There's a `;` in between. Can you make the other line to appear on the next line?

app/src/main/java/org/fossasia/phimpme/base/BaseActivity.java
```diff
@@ -132,13 +132,25 @@ public boolean onNavigationItemSelected(@NonNull final MenuItem item) {
         if (item.getItemId() != getNavigationMenuItemId()) {
             switch (item.getItemId()) {
                 case R.id.navigation_camera:
-                    startActivity(new Intent(this, CameraActivity.class).addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP));
+                    Intent cameraIntent = new Intent(this, CameraActivity.class);
+                    startActivity(cameraIntent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP));
+                    finish();
+                    overridePendingTransition(R.anim.left_to_right,
```
abishekvashok: Can finish and this line be moved out of the switch?

yashLadha:
Done @abishekvashok

raj10071997:
@yashLadha  I see that in some places you have initialized the variable outside the function calling and then passed them in the argument. Like this - 
` ` `
  Intent cameraIntent = new Intent(this, CameraActivity.class);
            startActivity(cameraIntent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP));
` ` `
Actually I don't know if this improves the performance. But in my view, java is pass by value, so if you initialize the variable inside the function call or outside the function call and then pass it as parameter, what difference does it create? Because whatever will be the case first the parameters will be resolved  and memory will be assigned to them first and then their memory reference will be passed as value. Actually I don't know whether writing the code like this will improve the performance but this was a genuine doubt of mine, so I asked the question?

yashLadha:
Actually the animation currently is sort of periodic and not according to the previous state. I am planning to change the logic in other PR about including the animation according to a specific logic. @raj10071997

PR #1918, opened 2018-03-02
Changing namespace from android to app

### LINKED ISSUE #1917: Apply method on SharedPreferences
**Actual Behaviour**

Shared preferences use the method commit() to store key value pair in PDK library. 

**Expected Behaviour**

SharedPreferences should use the method apply() instead of commit() to store key value pair  in PDK library.

**Steps to reproduce it**

Change the method from editor.commit() to editor.apply()

Yes, I would like to work on the issue.

DhruvamSharma:
Fixed #1917 

Changes: Change in the file (pdk/src/main/java/com/pinterest/android/pdk/PDKClient.java) in line 542 and 530


yashLadha:
Please change the PR description as well as the PR title in a more informative manner.

app/proguard-rules.pro
```diff
@@ -0,0 +1,342 @@
+# Add project specific ProGuard rules here.
```
yashLadha: Why `proguard-rules` is relevant to this PR??

pdk/src/main/java/com/pinterest/android/pdk/PDKClient.java
```diff
@@ -527,7 +527,7 @@ private void saveAccessToken(String accessToken) {
         SharedPreferences sharedPref = _context.getSharedPreferences(PDK_SHARED_PREF_FILE_KEY, Context.MODE_PRIVATE);
         SharedPreferences.Editor editor = sharedPref.edit();
         editor.putString(PDK_SHARED_PREF_TOKEN_KEY, accessToken);
-        editor.commit();
+        editor.apply();
```
yashLadha: It will be better to use `commit()` on a separate thread, apart from the UI thread. The reason is `commit()` is synchronous while `apply()` is asynchronous. So in case it might not perform actions immediately as expected.

DhruvamSharma: Should I create a separate thread for using commit?

yashLadha: Yes, or it would be better to create a generic class for the same.

DhruvamSharma: How should I achieve this?

yashLadha: Create a Java class that takes an object and call a function to execute the same on a separate thread.

abishekvashok:
Please do the changes relevant to the pr only in the pr. This contains work that you've done in earlier ones I guess. You can always use different branches :)

PR #1927, opened 2018-03-04
Fix #1824 Fixed delete option in captionDialogBox

### LINKED ISSUE #1824: functionality of delete option in captionDialogBox in sharingActivity is not same as description dialog box
**Actual Behaviour**

The delete option on captionDialogBox in SharingActivity.java does not do same job as the delete option in description dialog box.

**Expected Behaviour**

The delete option should behave exactly like the delete option in description dialog box as to avoid the confusion.

**Screenshots of the issue**

delete

Yes.

raj10071997:
Fixed #1824

Changes: Before in the Description Dialog box, the delete option was enabled only when there is saved description. The option is enabled when you open the description dialog box again after saving the description. The issue is that when you open the dialog box again and change the text (just change the text and do nothing) the delete option becomes disabled and the color of the delete option is set to the accent color which shows that the delete option is enabled but it isn't. Since delete option is an option to delete something from the database it should be available all the time even if the text changes. So made some to make delete option available all the time if there is some saved description in the database and added the same function in the delete option in captionDialogBox in ShareActivity.java.


app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -860,14 +864,11 @@ public void onClick(DialogInterface dialog, int which) {
...
             public void onClick(View v) {
-                String captionText = captionEditText.getText().toString();
-                if(!captionText.isEmpty()){
                     caption = null;
                     text_caption.setText(caption);
                     captionEditText.setText(text_caption.toString());
                     passwordDialog.getButton(DialogInterface.BUTTON_NEUTRAL).setEnabled(false);
-                }
-                passwordDialog.dismiss();
+                    passwordDialog.dismiss();
```
mayank8318: I think the indentation here is wrong.

raj10071997:
@isuruAb @abishekvashok @anantprsd5 Please review this PR.

sauravvishal8797:
@raj10071997 Please update the branch.

PR #1931, opened 2018-03-05
Fix #1922 - Add Edge detection as a separate filter.

### LINKED ISSUE #1922: Add Edge detection as a separate filter
**Actual Behaviour**

Currently, the edge detection filter is in the `default` case in `main_processing.cpp` file.

**Expected Behaviour**

It should be an independent filter.

Yes

mayank8318:
Fixed #1922 

Changes: Added edge detection as a separate filter named 'Edgify'

Screenshots of the change: 
image

image


mayank8318:
@isuruAb @anantprsd5 @abishekvashok Please review :)

PR #1947, opened 2018-03-10
Fix #1923-Moved Add to Favourites operation to a BackGround Task.

### LINKED ISSUE #1923: Adding multiple images to favourites issue.
**Actual Behaviour**

On adding large no of images at once(consider adding an entire folder containing about 20+ images) the UI freezes.

**Expected Behaviour**

The add to favourites task(multiple) can be moved to a background task, so that the UI doesn't freezes.

**Steps to reproduce it**

Try adding an entire folder to favourites at once.

Yes.

sauravvishal8797:
Fixed #1923 

Changes: Moved Add to Favourites operation to a BackGround Task.


app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -3217,6 +3180,64 @@ protected void onPostExecute(Void result) {
...
+        @Override
+        protected void onPreExecute() {
+            getNavigationBar();
+            swipeRefreshLayout.setRefreshing(true);
+            super.onPreExecute();
+        }
+
+        @Override protected Integer doInBackground(Void... voids) {
+            int count = 0;
+            realm = Realm.getDefaultInstance();
```
mayank8318: Isn't this already initialized?

sauravvishal8797: @mayank I guess realm objects can only be accessed on the thread they were created and this is a background thread so in order perform realm operation we need to get an instance of realm in this thread. Same goes for every other background thread using realm.

sauravvishal8797:
@abishekvashok @isuruAb Please review.

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
mayank8318: Why not use a local instance then?

raj10071997: I think @mayank8318 is right. See this link - https://stackoverflow.com/questions/407255/difference-between-declaring-variables-before-or-in-loop . The answer says that don't leave a narrow gap between declaration and initialization.

sauravvishal8797: @mayank8318 @raj10071997 Yeah that can be done, will make the change.

sauravvishal8797: @mayank8318 @raj10071997  I don't think this is necessary since the realm object has been used quite a few times in LFMainActivity for the purpose of operating with favourites part, so I guess it'd be better if a Realm object is declared at the top(like it is now) and every time a realm instantiation is required this object can be used.

sauravvishal8797:
@isuruAb @abishekvashok @anantprsd5 Please review this PR.

PR #1949, opened 2018-03-11
Fix #1944-Long press operation issue fixed, modified unfavourite func…

### LINKED ISSUE #1944: Bug in photos long press operation.
**Actual Behaviour**

On selecting multiple photos via the long press operation, in the overflow menu both option to Add to favourites as well as option to Unfavourite images are visible.

**Expected Behaviour**

Only add to favourites option should be visible for photos in the edit mode. Users can remove images from the favourites collection by navigating to the favourites section and using the Unfavourite option.

**Steps to reproduce it**

Select multiple photos through the long press operation, go to the overflow menu. 

**Screenshots of the issue**

screenshot_2018-03-09-18-36-57-406_org fossasia phimpme

Yes.

sauravvishal8797:
…tionality.

Fixed #1944 
Changes: Unfavourite option is now only visible in the favourites mode. Previously both add to favourites and Unfavourite option were visible on performing the long press operation, which was somewhat confusing so I guess it's better to limit the Unfavourite option to favourites section only(the user can navigate to the Fav section and perform the remove operation). Added a dialog to confirm the Unfavourite operation in the favourites section(this was there before Unfavourite option was implemented). Also modified the message displayed in the dialog to remove photos from favourites via the SingleMediaActivity.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
@@ -790,14 +790,20 @@ public class SingleMediaActivity extends SharedMediaActivity implements ImageAda
                 break;
 
             case R.id.action_delete:
+                String ButtonDelete = "";
                 handler.removeCallbacks(slideShowRunnable);
                 final AlertDialog.Builder deleteDialog = new AlertDialog.Builder(SingleMediaActivity.this, getDialogStyle());
-
-                AlertDialogsHelper.getTextDialog(SingleMediaActivity.this, deleteDialog,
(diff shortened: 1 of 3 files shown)


sauravvishal8797:
@abishekvashok Please review.

app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
```diff
@@ -790,14 +790,21 @@ public void folderSelected(String path) {
                 break;
 
             case R.id.action_delete:
+                String buttondelete = "";
```
mayank8318: Camel case :p

PR #1961, opened 2018-03-19
Fix #1960 - Caption is not shared to Instagram

### LINKED ISSUE #1960: Caption is not shared to Instagram
**Actual Behaviour**

While sharing the Image with caption to Instagram, Caption can not be shared as per Instagram Policy.

**Expected Behaviour**

User can be asked through the AlertDialog "Do you want to copy caption?", then user can paste the caption easily on Instagram.

Yes

ashley8jain:
Fixed #1960 

Screenshots of the change:


yashLadha:
The job of the user is not reduced at all.

ashley8jain:
If user share to instagram, then he forgets that caption cant be shared to instagram, so user have to go back to phimpe app to copy caption and then again share to instagram.
Moreover, copy the caption takes times (press the pointer -> "SELECT ALL -> "COPY").

yashLadha:
The intuitive thing user can also do to write in the Instagram app only
instead of phimpme, but user should write the caption in phimpme and that
has to be shared with instagram also. Hope you are getting the point.

ashley8jain:
User can also write the caption on every app like whatsapp, facebook. But why is there caption on phimpme app?
Therefore, caption should be shared from instagram. If user want to write the caption directly on instagram, then he can press "No" in dialog title "Do you want to copy?".

yashLadha:
I am also saying the same the app should share the caption and not the user
should copy paste the content, it seems like a clipboard managing app.

ashley8jain:
Yeah caption cant be shared to instagram as per instagram policy, so caption can be copied automatically for easy & quick convenience

yashLadha:
Them in that case there can be two choices:
1. Notify user that Instagram cannot share caption..and remove the text
field, because other apps are also following the same philosophy on play
store.
2. Update the message in alert dialog.
 If we think quantitatively how much work the user has to do them disabling
the edit text and displaying the toast of cannot share caption directly
will enable user to write only a single time, but in other case user has to
write and them paste the content also clicking on the alert dialog as well.
So number of actions are more in second option as compared to other. If we
think this way the number of action user have to perform for such a simple
task them it makes difference.

ashley8jain:
@anantprsd5 @isuruAb what do you think?

Issue #1911, opened 2018-02-28
Issue in Accounts section bottom bar.


sauravvishal8797:
**Actual Behaviour**

The bottom is displayed in a different colour than the primary colour.

**Expected Behaviour**

The bottom bar should be displayed in the primary colour.

**Steps to reproduce it**

Go to accounts section by taping on the icon on the LFMainActivity bottom bar.

**Would you like to work on the issue?**

Yes..

raj10071997:
@sauravvishal8797  can you check if the toggle button which sets the primary color of the navigation bar in the settings was on or off?

sauravvishal8797:
@raj10071997 The button was on.

raj10071997:
@sauravvishal8797  Can you please rewrite the steps to reproduce the issue as I am not able to reproduce the effects? In my app the Accounts activity tool bar and bottom bar both are being displayed according to primary color.

sauravvishal8797:
@raj10071997 Guess it was just a one-time problem, working fine for me too therefore I'm closing this issue.

PR #1974, opened 2018-03-30
Fixes : #1973

### LINKED ISSUE #1973: Change toolbar color in accounts activity
**Actual Behaviour**

Toolbar and navigation bar color are not matching

**Expected Behaviour**
They should match

Yes

yashLadha:
Fixed #1973 

Changes: Changed the top navigation bar to same color as bottom bar.

Screenshots of the change: 

screenshot


yashLadha:
Ping @abishekvashok @isuruAb

DhruvamSharma:
I think, what the normal scenario be is, that the bottomNavigationBar should have the blue color rather than purple

yashLadha:
When i compiled the latest bbuild the color of accounts is changes so i
reverted the top bar. If that's the case i will change it to it's previous
state. Views needed?

yashLadha:
Nope it's not.

raj10071997:
@yashLadha what color did you see for the navigation bar and the tool bar? Did you get navigation bar color- dark blue and the tool bar color according to the primary color?

yashLadha:
I have selected the same colour as the navigation bar in account activity.

raj10071997:
@yashLadha I am sorry, I didn't get your issue. Is your issue same as this one #1911 because I have also noticed this issue once or twice while using the app. If not, then in settings there is an option Colored Navigation Bar. If you switch off the toggle button then the bottom navigation bar color remain fixed according to previous primary color when the button was on. So if you change the primary color now, the bottom bar color doesn't change. But as soon you turn on that button the bottom bar color becomes same as primary color. So can you please confirm that your issue is same as #1911 ?

yashLadha:
@raj10071997 Yes it seems but i can't say for sure as there is no screenshot. For the time being why not change the color scheme for the account to be same as for the whole app.

DhruvamSharma:
I think it should be same for the whole app.

raj10071997:
@yashLadha @DhruvamSharma It is same for the whole application still when the toggle button is on. It is just that, before the toggle button did nothing, later this functionality was added to this button according to the name of the button "Colored Navigation Bar". So if you want to make the primary theme same for the whole application you can turn on toggle button and change the theme, it will be changed for the whole application. And I think that the issue faced by @yashLadha is because of the two methods ```getPrimaryColor()``` and ```getPrimaryColor(Context context)``` in ThemeHelper.java because the default color of the primary color returned from these two methods are different. That's why I asked about the color of the navigation bar, if it was dark blue or not. What do you think @yashLadha  @DhruvamSharma ?

raj10071997:
Yes, you are calling the same function for the navigation and the bottom bar, but what I don't understand that even after setting the primary color and saving the values in shared preferences, why I faced that issue while using the application. It might be because the shared preference returned the default value. Actually I don't understand the actual reason for the issue because both the functions are returning primary color.

PR #1976, opened 2018-03-30
Fixes #1975: Fixes Memory Leaks in LFMainActivity

### LINKED ISSUE #1975: Leak in AsyncTask in LFMainActivity
**Actual Behaviour**

There is a memory leak in AsyncTask in LFMainActivity.

**Expected Behaviour**

There should not be any memory leak in the code to maximise the performance.

**LogCat for the issue**

One of the AsyncTask Log:

```
This AsyncTask class should be static or leaks might occur (org.fossasia.phimpme.gallery.activities.LFMainActivity.PrepareAlbumTask) less... (Ctrl+F1) 
A static field will leak contexts.  Non-static inner classes have an implicit reference to their outer class. If that outer class is for example a Fragment or Activity, then this reference means that the long-running handler/loader/task will hold a reference to the activity which prevents it from getting garbage collected.  Similarly, direct field references to activities and fragments from these longer running instances can cause leaks.  ViewModel classes should never point to Views or non-application Contexts.
```

Yes

yashLadha:
Fixed #1975 
Changes: Removed leaks from the AsyncTasks in LFMainActivity


yashLadha:
Ping @abishekvashok @isuruAb

mayank8318:
Please add a proper PR title

yashLadha:
Done @mayank8318

yashLadha:
Yes, it is. Travis builds are failing because it is not able to download
some dependencies.

yashLadha:
Now the builds are fixed @mayank8318 . cc @abishekvashok @isuruAb

yashLadha:
Ping @mayank8318

PR #2031, opened 2018-06-13
Fix #2029-Made changes to provide box sharing feature only in devices…

### LINKED ISSUE #2029: Box will remove support for TLS 1.0
Box will remove support for TLS 1.0 on June 15, 2018. To keep your API communications secure, developers using the Box API are required to upgrade their applications to support TLS 1.1 or higher before this date.

To avoid any potential service disruptions:

Step 1: Test to see if your application(s) is correctly configured by following this guide available on our developer site: http://mkt.box.com/e11JKJ00020OCx0E9aQjV0s

Step 2: If the test fails, upgrade your application to support TLS 1.1 or higher following this guide: http://mkt.box.com/e11JKJ00020OCx0E9aQjV0s

sauravvishal8797:
… running version 5.0 and above.

Fixed #2029 

Changes: With Box removing support for TLS 1.0, TLS 1.1 will only be supported by Android versions 5.0 and above. Therefore I have made changes to enable the Box upload/share feature in the Accounts Sections only for devices running 5.0 and higher(so the app would still run on a device running on KitKat version but without the Box share feature).

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/accounts/AccountAdapter.java
+++ b/app/src/main/java/org/fossasia/phimpme/accounts/AccountAdapter.java
@@ -1,5 +1,6 @@
 package org.fossasia.phimpme.accounts;
 
+import android.os.Build;
 import android.support.v4.content.ContextCompat;
 import android.support.v7.widget.CardView;
 import android.support.v7.widget.RecyclerView;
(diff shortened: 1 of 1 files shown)


sauravvishal8797:
@mariobehling @anantprsd5 Please review.

PR #2046, opened 2018-06-19
Fix #1854-Move operation for favourite marked photos solved.

### LINKED ISSUE #1854: The star icon disappears and even the image from the favourites if the file is moved to some other path.
**Actual Behaviour**

If you change the path of an image (which is already marked favourite) or move the image to some other folder, the image disappears from the favourite and the icon also disappears.

**Expected Behaviour**

There should be a warning given to the user that all the already marked favourites image will be removed from favourites, or the images can still be in favourites and the images can be moved. The second option increases the complexity of the moving option.

Yes.

sauravvishal8797:
Fixed #1854 

Changes: Now on moving photo/photos from one folder to other, a check is performed whether the photos are present in the favorites collections or not, if present the photos along with their new paths are added to the favourites collection and the favourite object of the photos with their previous/old paths are deleted from the realm database.


app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -2804,6 +2805,46 @@ private void getdescriptionpaths(String patjs, String newpth){
         }
     }
 
+    private void checkforfavourites(String path, ArrayList<Media> selectedphotos){
```
anantprsd5: It will be good if you follow camel case convention for naming functions.

sauravvishal8797:
@anantprsd5 Did the changes, please review.

PR #2145, opened 2018-08-09
Fix #2144-Fixes upload history crash.

### LINKED ISSUE #2144: App crash in Upload History section.
**Actual Behaviour**

On deleting an image from the upload history section via the SingleMediaActivity and pressing back the app crashes.

**Expected Behaviour**

The app should not crash.

Yes.

sauravvishal8797:
Fixed #2144 
Changes: There is some error while using the DiffUtil class for populating the recyclerview adapter, therefore have switched to using the notifyDataSetChanged() till the DiffUtil issue is resolved. Will solve the DiffUtil issue via another separate PR.


abishekvashok:
Thanks

PR #2148, opened 2018-08-10
Fix #2147-Trashbin empty view issue resolved.

### LINKED ISSUE #2147: Issues in Trashbin section when empty.
**Actual Behaviour**

When there are no items present in the trashbin section, delete button and restore option are still visible.
App crash for refresh operation in the empty state.
Empty view not displayed for the suitable situation when images are removed from the bin section via the popup menu options provided in the item view.

**Expected Behaviour**

Delete and restore functionality should be hidden from the user when the trashbin section is empty.
App should not crash.
Empty view should be displayed properly.

Yes.

sauravvishal8797:
Fixed #2147 

Changes: 1. crash fixed for refresh operation in the empty state.
                 2. delete option and restore option visibility issue for empty state fixed.
                 3. empty view display enabled on deleting/restoring the image via the popup menu options 
                provided in the item view(only when no image present in the bin).


abishekvashok:
Thanks :)

PR #2156, opened 2018-08-14
Merging development into master

sauravvishal8797:
Changes: Adding the latest changes into master.


sauravvishal8797:
@abishekvashok Yes, I think we can do the update. Most of the open issues are either enhancements or features which can be included in the next release. As of now, I guess there are no very major or immediate issues in the app so I believe we should do the update on play store now.

abishekvashok:
Let's do it then :)

PR #2159, opened 2018-09-02
Fix for issue onSingleTap stops working 

sparrow007:
Fixed #[1681]

Changes: [I have changed ImageAdapter and SingleMediaActivity in order to solve the issue]

Problem : I think through my tests on codes at first time when a viewHolder created and bind the data to it by adapter then it listen for all of it's listener and when we swipe it then again it bind new data but when we go back adapter position is not updated so all the listener for old position is not working so when we go to 3 position it will not listen for position 1 and 2.


abishekvashok:
Pls don't break an existing functionality.

sparrow007:
don't get me wrong but Is my code breaking the existing functions of the application ??

PR #2161, opened 2018-09-10
Fixed #2160 App does not crash while navigating quickly between editing tabs 

### LINKED ISSUE #2160: App crashes while navigating hastily through image editing tabs
**Actual Behaviour**
Presently the app crashes if we navigate hastily between the tabs displayed at the bottom while editing the image.

**Expected Behaviour**

The app should not crash and keep responding to what the user clicks.

**Steps to reproduce it**
 1) Open the App, click a picture and select the editing icon.
 2) Now repeatedly press different tabs quickly.
 3) The app crashes after a few clicks(BUG).

Yes

codepoet2017390:
Fixed #2160 

Changes: While navigating through the tabs at the bottom of the editing activity a separate Async task is called for loading the ArrayList attached to the tab. This ArrayList is null in the beginning and is initialized only when the doInBackground() method is called leading to crashes (ArrayList.size() causes null pointer Exception). So i implemented the OnpreExecute() method for Async Tasks and initialized the list in this method.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/fragment/FrameFragment.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/fragment/FrameFragment.java
@@ -128,7 +128,6 @@ public class FrameFragment extends BaseEditFragment {
         @Override
         public viewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
             View view = View.inflate(getContext(), R.layout.frames, null);
-
             return new viewHolder(view);
         }
         @Override
--- a/app/src/main/java/org/fossasia/phimpme/editor/fragment/FrameFragment.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/fragment/FrameFragment.java
@@ -128,6 +128,7 @@ public class FrameFragment extends BaseEditFragment {
         @Override
         public viewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
             View view = View.inflate(getContext(), R.layout.frames, null);
+
             return new viewHolder(view);
         }
         @Override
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/editor/fragment/FrameFragment.java
```diff
@@ -128,6 +128,7 @@ private boolean checkVisibility() {
         @Override
         public viewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
             View view = View.inflate(getContext(), R.layout.frames, null);
+
```
sauravvishal8797: Please remove this extra space.

codepoet2017390:
I am new to Open Source so i didnt know about the space issue. I will surely take care in future.
Thankyou

app/src/main/java/org/fossasia/phimpme/editor/fragment/FrameFragment.java
```diff
@@ -298,6 +298,11 @@ public void setHeight(int height) {
...
     private class asyncThumbs extends AsyncTask<Void, Void, Void> {
 
+        @Override
+        protected void onPreExecute() {
+            arrayList=new ArrayList<>();
+        }
+
         @Override
         protected Void doInBackground(Void... params) {
             arrayList = new ArrayList<>();
```
sauravvishal8797: @codepoet2017390 You have already initialized this ArrayList in the onPreExecute method, then why do it here also?

sauravvishal8797: @codepoet2017390 You have already initialized this ArrayList in the onPreExecute method then why do it here too?

PR #2163, opened 2018-09-11
FIXED #2162 All GIF files are saved properly

### LINKED ISSUE #2162: Unable to create more than one gif at a time.
**Actual Behaviour**

Only one GIF file is found in Phimphme_gifs directory even after creating multiple gifs.

**Expected Behaviour**

All the GIF files must be saved separately in the Phimphme_gifs directory.

**Steps to reproduce it**

1. Open the app and go to any directory.
2. Select some pictures and create a gif ,repeat the same 2-3 times with different set of images.
3. Open the phimphme_gifs directory,It will show only one gif file.(BUG)

Yes

vinmaxx99:
Fixed #2162 

Changes: The GIF files were created with the same filename (GIF.gif) due to which the newly created gif was overwritten to the old gif and only one file was saved at a time ,so I added date and time to the gif filename so that the new files don't overwrite the previous ones.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -117,8 +117,10 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.ref.WeakReference;
+import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -3351,10 +3353,13 @@ public void run() {
             }
             byte[] bytes = createGIFFromImages(bitmaps);
             File file = new File(Environment.getExternalStorageDirectory() + "/" + "Phimpme_gifs");
+            DateFormat dateFormat = new SimpleDateFormat("dMMMyyyy_HHmmss");
```
abishekvashok: It should have been ddMMYY_HHmm
We don't want seconds :P

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -3351,10 +3353,13 @@ public void run() {
             }
             byte[] bytes = createGIFFromImages(bitmaps);
             File file = new File(Environment.getExternalStorageDirectory() + "/" + "Phimpme_gifs");
+            DateFormat dateFormat = new SimpleDateFormat("dMMyy_HHmm");
```
abishekvashok: It should be `dd` not `d`

vinmaxx99:
I am really sorry for the mistakes ,this is my first PR,I will make sure it never happens again in the future.

PR #2178, opened 2018-09-27
Fix #2177 List should be orientation dependent (Frame section)

### LINKED ISSUE #2177: List should be orientation dependent (Frame section)
**Actual Behaviour**

In Frame section of the edit Image does not change it's orientation means when the phone is in the portrait mode the frames list shows in horizontal which is fine but when changes to the landscape list still showing horizontally.

**Expected Behaviour**

List should be dependent on orientation, if it's in portrait list should shows horizontally and if it's in landscape it should be vertical.

yes

sparrow007:
Fixed #2177 

Changes: [When phone orientation changes layout should also changes it orientation so in order to do this i added a condition in FrameFragment for changing the orientation. ]

GIF of the change:


abishekvashok:
Thanks

PR #2190, opened 2018-10-02
Fix #2189 (Fragment overlapping on orientation change in the edit image section)

### LINKED ISSUE #2189: Fragment overlapping on orientation change in the edit image section.
**Actual Behaviour**

when selecting other section(fragment) like enhance(fragment) or any other and then change the orientation first which filter is overlap on enhance. 

**Expected Behaviour**

it should not overlap to another fragment.

**Steps to reproduce it**

Step 1. Click on edit image when switch from filter to enhance.

Step 2. Change the orientation of the phone (landscape mode) you will at filter fragment and just scroll the preview you will see these preview are overlapping on the switch fragment content.

**Screenshots of the issue**

whatsapp image 2018-10-03 at 3 15 04 am

yes

sparrow007:
Fixed #2189 

Changes: [transaction add() in fragment added the fragment on above of another so i used replace() to make only one fragment at time at the preview container]

Screenshots of the change: 

whatsapp image 2018-10-03 at 3 52 13 am

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/EditImageActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/EditImageActivity.java
@@ -174,7 +174,7 @@ public class EditImageActivity extends EditBaseActivity implements View.OnClickL
 
         getSupportFragmentManager()
                 .beginTransaction()
-                .add(R.id.preview_container, filterFragment)
+                .replace(R.id.preview_container, filterFragment)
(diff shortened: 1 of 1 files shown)


abishekvashok:
Thank you.

I see you are involved here. It would be better if you could follow up with better commit messages from the next time. Here's a guide: https://chris.beams.io/posts/git-commit/

Issue #2175, opened 2018-09-22
Edits not being saved when switching from portrait to landscape mode while editing an image


codepoet2017390:
**Actual Behaviour**

Currently the edits made and saved to an image are not being retained while switching from portrait to landscape mode 

**Expected Behaviour**
Edits made should be retained when the orientation is changed so that the user can edit the image further. 

**Steps to reproduce it**
  Step 1.
Select any image and click on the edit button.

Step 2.
 make changes in the image and keep saving each of them by clicking the tick-mark icon provided. 

Step 3.
change the orientation of you phone.

Step 4.
 The changes made are reset(BUG).

**Would you like to work on the issue?**
 YES

Divyanshu2132:
YES

codepoet2017390:
> YES

 Could you please clarify the context of your comment @Divyanshu2132

yashk2000:
This issue has been fixed by pr https://github.com/fossasia/phimpme-android/pull/2790

PR #2198, opened 2018-10-04
Fix #2194 app crashes in crop fragment depends on orientation

### LINKED ISSUE #2194: App crashes if the orientation is changed when click on the crop of the image
**Actual Behaviour**

App crashes if click on crop (Transform section) and then on orientation change.

**Expected Behaviour**

App should not crash.

**Steps to reproduce it**

Step 1. Click on edit button of the image and select the transform section.

Step 2. Click on crop button and changed the orientation without saving it then app will crash.

Yes

sparrow007:
Fixed #2194 

Changes: [crash due to calling the null bitmap because of changing the orientation so i apply few conditions in order to stop the app crashes.]

GIF  of the change:

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/fragment/CropFragment.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/fragment/CropFragment.java
@@ -77,7 +77,7 @@ public class CropFragment extends BaseEditFragment {
 	}
 
 	private void resetCropView() {
-		if (null != activity && null != mCropPanel){
+		if (null != activity && null != mCropPanel && activity.mainBitmap != null){
 			mCropPanel.setVisibility(View.GONE);
 			RectF r = activity.mainImage.getBitmapRect();
 			activity.mCropPanel.setCropRect(r);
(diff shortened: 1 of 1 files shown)


abishekvashok:
Can you make it work in a way that, the crop function still remains on the same state, instead of resetting?

sparrow007:
yeah i will try for that but i send this PR just to stop crash in the app.

sparrow007:
I am going to create new issue for that and work on it.

abishekvashok:
> I am going to create new issue for that and work on it.

Nope, please do it in this PR itself.

abishekvashok:
Updates?

sparrow007:
@abishekvashok  I am working on it , i will give you update soon.

sparrow007:
@abishekvashok  PR #2231 fix it (PR makes the cropFragment remain in it first position) although it's not remain in state which is changes in crop position in portrait mode will also remain same that is the case of this issue #2175 but it will appear crop rectangle after changes the orientation.

abishekvashok:
Please fix it completely.

codepoet2017390:
> I am ready with the fix for #2175 will put it up by tonight
> Please fix it completely.

@abishekvashok 
This is regarding Fix for Issue #2175  and the feature for retention of state when edit is not saved. I have fixed Issue #2175 and i'm halfway through to fix the other issue. Small bugs keep creeping in when i try to fix these issues as they are spread over a large number of activities. So i need some time and i assure a quality fix to it.

abishekvashok:
Take your own time

PR #2208, opened 2018-10-05
Fix #2201 (Fragment should remain same after orientation changed)

### LINKED ISSUE #2201: Fragment should remain same after orientation changed
**Actual Behaviour**

When we edit the image it shows different options (Fragments) and when we select one of them and change the orientation of phone it switching to first option which is filter.

**Expected Behaviour**

Selected fragment should be same even after changed the orientation of the phone.

**Steps to reproduce it**

Step 1. Click on the edit image button and select different options like Frame.

Step 2. Change the orientation of the phone it goes to first option (Fragment) which is filter.

yes

sparrow007:
Fixed #2201 

Changes: [Traditionally savedInstanceState() used to restore the state of activity so implement it in EditImageActivity to restore selected fragments.]

GIF of the change:


abishekvashok:
Well done. Thanks

PR #2231, opened 2018-10-11
Fix #2209 ContainerFragment in EditImageActivity changes it's layout on orientation change

### LINKED ISSUE #2209: ContainerFragment in EditImageActivity changes it's layout on orientation change
**Actual Behaviour**

On selecting sub component of in container fragment and then changing the orientation change the layout.

**Expected Behaviour**

layout should not be changed.

**Steps to reproduce it**

Step 1. Click on edit image button and click any of option like enhance option

Step 2. click on contrast then it shows a progress bar

Step 3 Change the orientation of the phone, layout changes.

**Screenshots of the issue**

whatsapp image 2018-10-05 at 11 56 35 pm

yes

sparrow007:
Fixed #2209 

Changes: [There are different issues in the EditImageActivity due to the orientation either it is app crash or it may differ the layout changes so what i added to prevent it from app crash as well as layout changes , i added two variable to handle two different fragment.
1. middleMode 
2. bottomMode 
These two will  handle all the fragment in the EditImageActivity, previously EditImageActivity  has only one variable name mode which handles both work so many times changes in mode affect both fragments in the EditImageActivity. ]

Screenshots of the change: 

whatsapp image 2018-10-12 at 1 40 14 am

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/EditImageActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/EditImageActivity.java
@@ -87,6 +87,9 @@ public class EditImageActivity extends EditBaseActivity implements View.OnClickL
     public static int mode;
     public static int effectType;
 
+    public static int middleMode = MODE_FILTERS;
+    public static int bottomMode = MODE_MAIN;
+
     /**
      * Number of times image has been edited. Indicates whether image has been edited or not.
      */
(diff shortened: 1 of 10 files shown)


sparrow007:
would you please explain more?

abishekvashok:
I vote against the solution provided.

> There are different issues in the EditImageActivity due to the orientation either it is app crash or it may differ the layout changes so what i added to prevent it from app crash as well as layout changes , i added two variable to handle two different fragment.
middleMode
bottomMode
These two will handle all the fragment in the EditImageActivity, previously EditImageActivity has only one variable name mode which handles both work so many times changes in mode affect both fragments in the EditImageActivity. 

Instead of doing this, please inspect the real reasons of the crash by inspecting logs, and fix it as such. Making two variables to store fragment IDs don't make sense, as in a fragment, the bars should be in sync with it. The current solution may upset the sync. I can agree that this would fix the issues, but we need `maintainable code`, and in my prespective this doesn't fit into the definition of such code.

You can always write better code than this, and please do it. Take time and enjoy the process instead of just trying to hack it to a fix.

mariobehling:
@sparrow007 I am closing this due to inactivity. Please feel free to provide a new PR once you are ready. Thank you!

PR #2236, opened 2018-10-21
Fixed#2235 Zip Issue Fixed

### LINKED ISSUE #2235: Unable to make Zip file for images with .jpeg and .png extensions
**Actual Behaviour**

Presently the app does not zip files ending with extensions .jpeg and .png.

**Expected Behaviour**
The user should be able to zip files with extensions .jpeg and .png

**Steps to reproduce it**
1) Select an album which contains files with file extensions (.jpeg), (.png) and (.jpg)  
2)  Select the Zip option from the overflow menu by clicking in the top right corner 
3) You will see that the Zip is only created for files with extension .jpg

YES

codepoet2017390:
Fixed #2235 

Changes: Only jpeg files were being added to path so i improved the if condition which checked which files go into the path


abishekvashok:
We could have done written better code for this, but currently lets improve it latter :)

abishekvashok:
@codepoet2017390 thanks for working on it

PR #2243, opened 2018-10-23
Fixed#2239 Used setcomponent instead of setpackage for Snapchat

### LINKED ISSUE #2239: Image is lost when sharing via Snapchat
**Actual Behaviour**
When we share an image via Snapchat in the share image activity. Snapchat fires up with its Main Activity and the image is lost.

**Expected Behaviour**
The share image activity of Snapchat should open up with our image preloaded in it. 

**Steps to reproduce it**
1) Open an image and press the share icon.
2) Now press the Snapchat icon displayed there.
3) Snapchat fires up and our image is lost.

   YES

codepoet2017390:
Fixed #2239 
Changes: Presently the Snapchat function used intent.setpackage() function which was however unable to open the required activity in snapchat so i replaced it with setComponent().


sauravvishal8797:
@codepoet2017390 Please resolve the conflicts.

PR #2269, opened 2018-11-06
Fixed#2255 App now accesses the previous password when password is set up again

### LINKED ISSUE #2255: Access by password should remember the password if it has been setup early
**Actual Behaviour**

The Security feature requires a password to be set up and it asks to set it up again and again.

**Expected Behaviour**

The app should remember the password and not ask for it again and again

**Steps to reproduce it**
1) Go to app settings and then go to security. 
2) Set up a password and then uncheck the Access by password.
3) Se it up again. It asks again for password.

YES

codepoet2017390:
Fixed #2255
Changes: App now accesses the previous password when password is set up again. Also the user can  always change the password if he wants to.


abishekvashok:
Thanks

PR #2275, opened 2018-11-10
Fixed #2226 Actionbar hidden on second screen when scrolls down in first screen 

### LINKED ISSUE #2226: Actionbar hidden on second screen when scrolls down in first screen
**Actual Behaviour**

when scroll down in album section action bar will  hidden then tap on the album and go to all the image of tapped album where action bar already hidden without scrolling.

**Expected Behaviour**

It should not hidden first, it should hide after scrolling the images.

**Steps to reproduce it**

Step 1. Scrolls down in album section when action will hidden.

Step 2. At that moment click on the album, it will open new page where all images of that album appears.

Step At first hidden action bar is still hidden

**Screenshots of the issue**

Yes.

sparrow007:
Fixed #2226 

Changes: [Added a new method for showing the toolbar when it goes from album to photos screen]

GIF of the change:

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -617,6 +617,7 @@ public class LFMainActivity extends SharedMediaActivity {
         public void onClick(View v) {
             fromOnClick = true;
             final Album album = (Album) v.findViewById(R.id.album_name).getTag();
+            showAppBar();
             //int index = Integer.parseInt(v.findViewById(R.id.album_name).getTag().toString());
             if (editMode) {
                 albumsAdapter.notifyItemChanged(getAlbums().toggleSelectAlbum(album));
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -715,6 +716,14 @@ private void clearOverlay() {
                 | AppBarLayout.LayoutParams.SCROLL_FLAG_ENTER_ALWAYS);
     }
 
+    //To show the appbar
+    private void showAppBar() {
+        if (toolbar.getParent() instanceof AppBarLayout) {
+            ((AppBarLayout)toolbar.getParent()).setExpanded(true, true);
+        }
+
```
codepoet2017390: Please remove this extra space and also i think there is no need of the comment as the method name is very much self-explanatory

abishekvashok:
Thanks

PR #2291, opened 2018-11-24
Fix for #2289 :A message is shown when password is removed

### LINKED ISSUE #2289: No message shown when password is removed
**Actual Behaviour**

If a present password is removed, no message is shown.

**Expected Behaviour**

When password is removed, a message should be given to the user indicating that there is no password set currently.

**Steps to reproduce it**

1) Go to phimp.me settings.
2) Enter the Security menu.
3) Set a password if not already done.
4) Make the "access by password" switch off. No notification is shown.

Yes

yashk2000:
Fixed #2289 

Changes: When a set password is removed, a message is shown to the user indicating that the password has been removed.

Gif of the change: 
password


yashk2000:
@codepoet2017390 my aim is not to remind the user that he has removed the password. I just want to notify the user the password has really been removed.

PR #2316, opened 2018-12-18
Fix for issue #2157 : Forgot password feature added.

### LINKED ISSUE #2157: Security feature enhancement.
**Actual Behaviour**

At the moment there is no forget password feature, so if a user forget's the password he'd need to reset all the settings to access the secure functionality again.

**Expected Behaviour**

There should be a forget password feature that would allow the user to reset the password without having to reset all the settings.

Yes.

yashk2000:
Fixed #2157 

Changes: Added a forgot password feature using security question to the app. 

GIF of the change: 

forgotpassword


sauravvishal8797:
@yashk2000 See if you could break the methods into smaller sub-methods.

PR #2332, opened 2018-12-25
Fix#2331

### LINKED ISSUE #2331: Account Section returns no response when internet is off.
**Actual Behaviour**

Accounts Activity gives no response when user is not connected to the internet

**Expected Behaviour**

A toast should be displayed when not connected to the internet.

**Steps to reproduce it**

1. turn off internet connection.
2. navigate to accounts activity.
3. no response when internet is off.

**Screenshots of the issue**

whatsapp image 2018-12-25 at 16 40 16

Yes.

kk63852:
Fixed #2331

Changes: 
When navigating through the accounts section a message is displayed if not connected to the internet.


abishekvashok:
Thanks

PR #2352, opened 2018-12-31
slideshow maximum duration was set to 10 sec

### LINKED ISSUE #2351: slideshow feature has no time limit
**Actual Behaviour**

slideshow feature has no time limit and the app crashes if the value is too high.

**Expected Behaviour**

there should be a maximum time limit.

**Steps to reproduce it**

open a image and navigate to menu options. go to slideshow feature and enter a very large value, the app would crash.

yes.

kk63852:
Fixed #2351 

Changes: 
maximum time duration was set to 10 sec and a toast was added.


abishekvashok:
Thanks

PR #2356, opened 2019-01-01
Fix for issue#2354 : Password length can now be upto 128 characters.

### LINKED ISSUE #2354: Password limit is only 10 characters.
**Actual Behaviour**

The maximum length of the password that can be entered by the user is 10 characters.

**Expected Behaviour**

The maximum length of password should be 128 characters according to [owasp](https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Password_Length).

**Steps to reproduce it**

1) Open security settings to set password.
2) The maximum length of password is 10 characters.

YES.

yashk2000:
Fixed #2354 

Changes: Passwords now have upto 128 characters.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
@@ -65,7 +65,6 @@ public class SecurityActivity extends ThemedActivity {
     public ArrayList<Album> albums;
     public ArrayList<String> securedfol = new ArrayList<>();
 
-
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -232,6 +231,7 @@ public class SecurityActivity extends ThemedActivity {
 
     private void setPasswordDialog() {
 
+        final short max_password_length = 128;
         final AlertDialog.Builder passwordDialog = new AlertDialog.Builder(SecurityActivity.this, getDialogStyle());
         final View PasswordDialogLayout = getLayoutInflater().inflate(R.layout.dialog_set_password, null);
         final TextView passwordDialogTitle = (TextView) PasswordDialogLayout.findViewById(R.id.password_dialog_title);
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
```diff
@@ -270,9 +270,9 @@ private void setPasswordDialog() {
             }
 
             @Override public void afterTextChanged(Editable editable) {
-                if(editable.length() == 11) {
-                    editTextConfirmPassword.setText(editable.toString().substring(0, 10));
-                    editTextConfirmPassword.setSelection(10);
+                if(editable.length() == 128) {
```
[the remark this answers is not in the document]
yashk2000: Right now if the length is greater than 128 then a toast is shown and any extra characters typed by the user are not shown. Should I still change it to greater than?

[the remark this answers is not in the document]
yashk2000: @abishekvashok this is what happens right now if 128 characters are reached:

different

Please tell whether I should make changes or not?

frextrite: This checks if the total characters entered are exactly 128, it then shows a pop-up displaying maximum password length. But what if someone pastes a password which is greater than 128 characters, then this statement would never be called, and the toast would never be shown.

yashk2000: @frextrite If someone pastes a password which is more than 128 characters long, the result would be the same. A toast is displayed and only the first 128 characters of the pasted text are shown.

yashk2000:
@frextrite I would have used a variable if the maximum length of the password was varying. I have fixed the maximum length to be 128 characters which cannot be varied so there isn't a need to use a variable.

app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
frextrite: That's great then!

frextrite:
Although you are right that max length value is fixed at ```128``` and won't change at runtime, but it's just good practice to use a variable, declare its value once and use it everywhere. It makes the code more readable, and easy to mantain. In this case, a ```const``` variable would do the job.

yashk2000:
@frextrite that's right. I'll make the change.

abishekvashok:
Thanks

PR #2367, opened 2019-01-08
Fix for issue#2365 : A password is asked to view all photos if security is active.

### LINKED ISSUE #2365: Even if password is on, all photos can be viewed.
**Actual Behaviour**

When the password is on, all photos can still be viewed.

**Expected Behaviour**

All photos should ask for the password before showing the photos.

**Steps to reproduce it**

1) Set a password if not set already. Lock local folders.
2) Go back to local folders. Tap on all photos. All photos including the locked photos can be viewed.

YES.

yashk2000:
Fixed #2365 

Changes: If security is on for local folders, the password will be asked to view all photos.

Screenshots of the change: 
screencap
screenc1ap


abishekvashok:
Thanks

PR #2393, opened 2019-01-15
After resetting password forgot password dialog is dismissed.

### LINKED ISSUE #2392: Forgot password dialog issue.
**Actual Behaviour**

When the user resets the password by pressing forgot password, after password reset, the user should directly be shown the insert password screen. But instead, the forgot password dialog appears again which has to be canceled to proceed to the insert password screen.

**Expected Behaviour**

The user should directly be shown the insert password screen.

**Steps to reproduce it**

1) Go to security settings to set a password if not set already.
2) Press fo forgot password when entering the password again.
3) After resetting the password, forgot password dialog is still open which has to be canceled to proceed to the inset password screen.

**GIF of the issue**

ezgif com-video-to-gif 1

YES.

yashk2000:
Fixed #2392 

Changes: After resetting the password, the user is directly shown the insert password screen instead of the forgot password dialog screen.

GIF of the change: 
ezgif com-video-to-gif 2

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/util/SecurityHelper.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/util/SecurityHelper.java
@@ -203,6 +203,7 @@ public class SecurityHelper {
                         answer1 = SP.getString("Security Answer", answer1);
                         if (Objects.equals(answer1, securityAnswer1.getText().toString())) {
                             changePassword(activity, passwordDialog);
+                            dialog.dismiss();
                         } else {
                             securityAnswer1.getText().clear();
                             til.setError(activity.getString(R.string.wrong_answer));


app/src/main/java/org/fossasia/phimpme/gallery/util/SecurityHelper.java
```diff
@@ -202,6 +202,7 @@ public void onClick(View view) {
                         String answer1 = "";
                         answer1 = SP.getString("Security Answer", answer1);
                         if (Objects.equals(answer1, securityAnswer1.getText().toString())) {
+                            dialog.dismiss();
                             changePassword(activity, passwordDialog);
```
abishekvashok: Move this above

yashk2000: @abishekvashok can you please tell me what do I have to move above? I did not understand what is to be moved.

abishekvashok: Dismiss the dialog only after the changepassword function is called.

yashk2000:
@abishekvashok I made the changes.

abishekvashok:
Thanks

PR #2400, opened 2019-01-16
Fix #2399 - Remove redundant drawable ic_search_black_24dp

### LINKED ISSUE #2399: Remove redundant drawable ic_search_black_24dp
**Actual Behaviour**

Drawable file ic_search_black_24dp.xml is currently not being used and hence is redundant.

**Expected Behaviour**

Drawable file ic_search_black_24dp.xml should not be present.

**Steps to reproduce it**

Delete ic_search_black_24dp.xml file

Yes, I would like to work on the issue.

Princejhabak:
Fixed #2399 

Changes: Unused drawable file ic_search_black_24dp.xml is deleted.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -215,7 +215,7 @@
         <!-- Close button icon -->
         <item name="closeIcon">@drawable/ic_close_black_24dp</item>
         <!-- Search button icon -->
-        <item name="searchIcon">@drawable/ic_search_black_24dp</item>
+        <item name="searchIcon">@drawable/ic_search_white_24dp</item>
         <!-- Layout for query suggestion rows // unused for now, staying with the standard layout -->
         <!--<item name="suggestionRowLayout">...</item>-->
     </style>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -215,7 +215,7 @@
         <!-- Close button icon -->
         <item name="closeIcon">@drawable/ic_close_black_24dp</item>
         <!-- Search button icon -->
-        <item name="searchIcon">@drawable/ic_search_black_24dp</item>
+        <item name="searchIcon">@drawable/ic_search_white_24dp</item>
```
abishekvashok: Err why did you remove it

Princejhabak: @abishekvashok, sorry but you are addressing the drawable file or styles.xml

abishekvashok: Styles.xml

Princejhabak: @abishekvashok, if the drawable file ic_search_black_24dp is removed then referencing it in styles.xml will produce error.  And drawable file ic_search_black_24dp must be deleted as it is currently not used anywhere in the project.

abishekvashok: The drawable was being used. You removed it's use and deleted it. See the entire diff

Princejhabak: @abishekvashok, I had to remove the use of the drawable file because of this

screenshot_20190119-102052


So I had to create ic_search_white_24dp

PR #2401, opened 2019-01-17
Discrete Rotation 90,180,270,360

### LINKED ISSUE #2097: 90 degree rotation
**Actual Behaviour**

You have to drag to set a rotation

**Expected Behaviour**

Rotate exactly 90 degrees losslessly (this is something you can do by editing the jpeg metadata)

anirudh0228:
This adds the feature of Discrete Rotation .A seekbar is used and the
image is rotated by 4 different angles(90,180,270,360) depending on the
value of the seekbar's progress.A seekbarchangelistener is used to
rotate the image and change the value of the horizontalWheelView
progress

Fixed #2097

Changes: RotateFragment.java,fragment_edit_image_rotate.xml

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/fragment/RotateFragment.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/fragment/RotateFragment.java
@@ -110,12 +110,12 @@ public class RotateFragment extends BaseEditFragment {
 
             @Override
             public void onStartTrackingTouch(SeekBar seekBar) {
-
+                //interface methods
             }
 
             @Override
             public void onStopTrackingTouch(SeekBar seekBar) {
-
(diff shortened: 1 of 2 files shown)


angmas1:
@anirudh0228 this adds discrete rotation feature, but removes the feature to allow any rotation. I think we should add this feature along with the feature which allowed any rotation.

anirudh0228:
@angmas1 thanks for your review. Both the features are working on my device . The SeekBar performs the discrete rotation and the HorizontalWheel performs the continous rotation . Maybe it's because of different devices .It would be really helpful if you can provide me the mobile model ,it will make the debugging process faster.
ezgif com-video-to-gif

abishekvashok:
Closing as obsolete

PR #2408, opened 2019-01-17
Removed obsolete configurations

### LINKED ISSUE #2405: Removing obsolete configurations
<b>Actual Behaviour</b>

Few configurations such as "compile" , "androidTestCompile","debugCompile" ,"releaseCompile" , "testCompile" have become obsolete now, but are still present in build.gradle.

<b>Expected Behaviour</b>

These configurations should be replaced with their new names , such as "compile" should be replaced with "implementation" , 'androidTestCompile' should be replaced with 'androidTestImplementation'.

<b>Would you like to work on the issue?</b>

Yes

Priyanka488:
Fixed #2405 

Changes: app/build.gradle , build.gradle , openCVLibrary24133/src/main/AndroidManifest.xml .
Changed 'compile' to 'implementation'
               'androidTestCompile' to androidTestImplementation'
               'debugCompile' to 'debugImplementation'
            'releaseCompile' to 'releaseImplementation'

### DIFF
--- a/openCVLibrary24133/src/main/AndroidManifest.xml
+++ b/openCVLibrary24133/src/main/AndroidManifest.xml
@@ -4,5 +4,5 @@
       android:versionCode="24133"
       android:versionName="2.4.13.3">
 
-    <uses-sdk android:minSdkVersion="8" />
+
 </manifest>
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -49,7 +49,7 @@ android {
     }
 
     configurations {
-        compile.exclude group: 'org.jetbrains', module: 'annotations'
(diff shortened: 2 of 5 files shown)


openCVLibrary24133/src/main/AndroidManifest.xml
```diff
@@ -4,5 +4,5 @@
       android:versionCode="24133"
       android:versionName="2.4.13.3">
 
-    <uses-sdk android:minSdkVersion="8" />
```
abishekvashok: Why was this removed?

Priyanka488: Actually there were some build errors so had to remove this. I'll try doing it again.

PR #2432, opened 2019-01-20
snackbar for no internet connection added

### LINKED ISSUE #2431: sharing activity has no internet check.
**Actual Behaviour**

if the internet is turned off while sharing an image, nothing is displayed.

**Expected Behaviour**

a snackbar is needed when the internet is off while sharing.

**Steps to reproduce it**

open a image and navigate to share option when the internet is turned off. tap on any sharing feature nothing is returned. 

yes.

kk63852:
Fixed #2431 

Changes:
a snackbar is now displayed if the internet is turned off and the user is sharing an image.

Screenshots of the change:


yashk2000:
@kk63852 this is not really required as user might also share images by offline means like messaging, shareIt etc.

kk63852:
this pr solves the message displayed issue. the building of the is not affected at all. its the same as before . you can compare this commit with the original code the result will be the same except snackbar will be shown.

yashk2000:
@kk63852 yes this is better. Just replace the snackbar with a toast message saying "No internet connection. Some apps might not behave as expected." This way even if the user shifts the activity, user can see the toast and know that he cannot use apps which require an internet connection.

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -281,107 +283,111 @@ public void onClick(View view) {
...
-        positionShareOption = position;
-        if (sharableAccountsList.get(position) == OTHERS) {
-            triedUploading = true;
-            shareToOthers();
-            return;
-        }
+        ConnectivityManager cm = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo ni = cm.getActiveNetworkInfo();
+        if (ni == null) {
+            Snackbar.make(findViewById(android.R.id.content), getString(R.string.internet_is_off), Snackbar.LENGTH_SHORT).show();
```
yashk2000: This snackbar should be replaced with a toast giving an appropriate message with LENGTH_LONG.

app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -225,7 +225,8 @@ public void onItemClick(final View childView, final int position) {
         final String name = AccountDatabase.AccountName.values()[position].toString();
 
         if (!signInSignOut.isChecked()) {
-            if (!checkNetwork(this, parentLayout)) return;
+            if(ni!=null)
+//            if (!checkNetwork(this, parentLayout)) return;
```
yashk2000: @kk63852 what is the reason for this change?

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -281,107 +283,111 @@ public void onClick(View view) {
...
     public void onItemClick(View childView, final int position) {
-        if (!checkNetwork(this, parent)) return;
-
-        positionShareOption = position;
-        if (sharableAccountsList.get(position) == OTHERS) {
-            triedUploading = true;
-            shareToOthers();
-            return;
-        }
+        ConnectivityManager cm = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
```
yashk2000: Instead of doing this, you could have used the function which is already created, checkNetwork().

app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
```diff
@@ -163,13 +163,13 @@ public static String getMimeType(String url) {
...
         return type;
     }
-    public static boolean checkNetwork(Context context, @NonNull View view) {
-        if (isInternetOn(context)) {
-            return true;
-        } else
-            SnackBarHandler.show(view,context.getString(R.string.not_connected));
-        return false;
-    }
+//    public static boolean checkNetwork(Context context, @NonNull View view) {
```
yashk2000: @kk63852 instead of creating your own method, use this method which is in utils. It has been created for the purpose of being called from anywhere. Or is there any particular reason for this change you made?

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
kk63852: well the snackbar what not working with this code. so i changed it. then i saw the usage of check network function, other than this, it was being used in accounts activity only for checking the network. connectivity manager was already setup so i changed it there. therefore it was not needed anymore.

yashk2000: There is no need. In the original code, in the if(!networkCheck()) return; is there. Instead of the return, create a toast there. I tried it, it's working.

yashk2000:
@kk63852 yes. This is fine. But in future please make changes only to the required code instead of aligning all the code. It makes too many unnecessary changes, like this is just a 6-7 line change, you've made a 71 line change.

yashk2000:
#2446

kk63852:
It was not done intentionally. It happened because a else if statement with braces was added, it was done by the studio. @yashk2000

PR #2445, opened 2019-01-22
redundant snackbar removed

### LINKED ISSUE #2444: alert and snackbar both appear after wrong password is given as input
**Actual Behaviour**

whenever wrong password is entered, alert in the dialog and snackbar in the activity both appear. 

**Expected Behaviour**

alert is enough. snackbar should be removed.

**Steps to reproduce it**

go to settings and set up security. (for example) open a album and press delete. input wrong password. a alert is shown in the dialog and a snackbar is also shown in the activity when the text-input is hidden. 

yes.

kk63852:
Fixed #2444

Changes: 
redundant snackbar removed.
some hard-coded strings were added in strings.xml


abishekvashok:
Thanks

PR #2450, opened 2019-01-24
forgot password dialog button change as per accent color

### LINKED ISSUE #2449: forgot password dialog button color issue
**Actual Behaviour**

forgot password button color not set according to accent color.

**Expected Behaviour**

it should change as per the accent color.

**Steps to reproduce it**

open forgot password dialog. the buttons are seen to be gray irrespective of the accent color change.

yes

kk63852:
Fixed #2449 

Changes: 
dialog button color set to accent color.
a hard-corded string was added to strings.xml


abishekvashok:
Next time make the change on my for one issue per pr

abishekvashok:
Thanks

PR #2459, opened 2019-01-28
fix #2458

### LINKED ISSUE #2458: Imagebutton for text color selection color issue.
**Actual Behaviour**

mTextColorSelector of edit text fragment background color is set to last line color picker even on cancelling the dialog.

**Expected Behaviour**

the color of the imagebutton should be set to white when the dialog is cancelled as the text color is set to white on cancelling and it should be set to whatever color picked from the line color picker when ok is pressed.

**Steps to reproduce it**

open a image and navigate to edit mode. go to write section and open text segment. tap on the imagebutton to open the line color dialog. change the color and cancel the dialog, the imagebutton color is set to whatever the line color picker color was set to before cancelling.

yes.

kk63852:
Fixed #2458

Changes: 
the imagebutton color now resets to white on cancelling the dialog.


abishekvashok:
Thanks

PR #2460, opened 2019-01-28
Fix for issue#2457: Fixed the view which was not looking good.

### LINKED ISSUE #2457: Appearance of the accounts section.
**Actual Behaviour**

There is a bar above the app bottom bar which does not look good in the accounts section.

**Expected Behaviour**

This bar should not be present.

**Steps to reproduce it**

Switch to accounts section, there is a bar present above the bottom bar of the app.

**Screenshots of the issue**

screencap

YES.

yashk2000:
Fixed #2457 

Changes: The bar which was showing up unnecessarily has been removed.

Screenshots of the change: 

screencap

screencap

screencap

### DIFF
--- a/app/src/main/res/layout/accounts_item_view.xml
+++ b/app/src/main/res/layout/accounts_item_view.xml
@@ -17,6 +17,7 @@
         card_view:cardBackgroundColor="@color/white"
         android:theme="@style/CardView.Light"
         android:layout_marginTop="3dp"
+        android:layout_marginBottom="5dp"
         android:elevation="3dp"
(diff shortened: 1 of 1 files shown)


yashk2000:
@kk63852 I tried doing it such that the gap would be minimum. I found no other way to remove that bar which was popping up without adding any padding. 
If there is a better way, please tell me, I'll make the changes in the pr.

kk63852:
the accounts are setup in a recycler view. that is why it seems difficult to remove the white lines.

PR #2462, opened 2019-01-28
Fix for issue#2461: Replaced reset settings snackbar with a toast.

### LINKED ISSUE #2461: Reset settings has a snackbar.
**Actual Behaviour**

Currently, reset settings has a snackbar to be displayed when settings are reset.

**Expected Behaviour**

Instead of a snackbar, there should be a toast as snackbars disappear when activity is shifted. Hence no message is displayed when settings are reset.

**Steps to reproduce it**

Go to reset settings. No message is displayed.

YES.

yashk2000:
Fixed #2461 

Changes: Replaced the reset settings snackbar with a toast so that it is always visible.

GIF of the change: 

ezgif com-video-to-gif 9


abishekvashok:
Thanks

PR #2464, opened 2019-01-29
fix #2463: multiple snackbar issue fixed.

### LINKED ISSUE #2463: multiple snackbar issue
**Actual Behaviour**

redundant snackbar in accounts activity.
snackbar is not displayed when the user is sharing an image personally on imgur and he is not signed in.

**Expected Behaviour**

redundant snackbar should be removed.
toast should be displayed instead of snackbar.

**Steps to reproduce it**

share a image via imgur when not signed in. tap on personal. no message is displayed, asking the user to sign in first.

yes.

kk63852:
Fixed #2463 

Changes: 
redundant snackbar from accounts activity removed.
snackbar of shareToImgur replaced with a toast.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
@@ -308,9 +308,6 @@ public class AccountActivity extends ThemedActivity implements AccountContract.V
                 /*case ONEDRIVE:
                     signInOneDrive();
                     break;*/
-
-                default:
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -308,9 +308,7 @@ public void onClick(View view) {
                 /*case ONEDRIVE:
                     signInOneDrive();
                     break;*/
-
-                default:
```
abishekvashok: Please don't leave an empty default

abishekvashok:
Thanks

PR #2466, opened 2019-01-29
fix #2465

### LINKED ISSUE #2465: accounts activity has a separate network check instead of using the function from utils
**Actual Behaviour**

accounts activity has a separate network check instead of using the function from utils.

**Expected Behaviour**

the function from utils.java should be used.

**Steps to reproduce it**

open accountsactivity.java and navigate to onItemClick. a separate network check instead of using the function from utils class.

yes

kk63852:
Fixed #2465 

Changes: 
separate network check was replaced with the function from utils class.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
+++ b/app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
@@ -19,10 +19,9 @@ import android.support.design.widget.Snackbar;
 import android.util.Base64;
 import android.view.View;
 import android.webkit.MimeTypeMap;
-import android.widget.Toast;
+
 
 import org.fossasia.phimpme.R;
-import org.fossasia.phimpme.accounts.AccountActivity;
 import org.fossasia.phimpme.data.local.AccountDatabase;
 
 import java.io.ByteArrayOutputStream;
--- a/app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
+++ b/app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
@@ -15,12 +15,14 @@ import android.net.Uri;
 import android.provider.MediaStore;
 import android.speech.RecognizerIntent;
 import android.support.annotation.NonNull;
+import android.support.design.widget.Snackbar;
 import android.util.Base64;
 import android.view.View;
 import android.webkit.MimeTypeMap;
 import android.widget.Toast;
 
 import org.fossasia.phimpme.R;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
```diff
@@ -167,7 +169,7 @@ public static boolean checkNetwork(Context context, @NonNull View view) {
         if (isInternetOn(context)) {
             return true;
         } else
-            SnackBarHandler.show(view,context.getString(R.string.not_connected));
+            Toast.makeText(context,R.string.internet_is_off,Toast.LENGTH_SHORT).show();
```
yashk2000: Why was this SnackBar changed to a toast?

kk63852: because this snackbar was not showing when checknetwork returned false.

yashk2000:
> Pls don't remove the snack bar and instead try to make it appear. Please check where the bug is

@abishekvashok I have opened #2467 and #2470 to deal with the snackbar position and visibility issues and am working on them.

kk63852:
@yashk2000 please exclude this snackbar from your pr since i have already opened the pr, i would resolve this snackbar issue.

yashk2000:
@kk63852 the change I will make might affect a lot of snackbars. But I will try my best to leave any changes pertaining to this snackbar.

kk63852:
@yashk2000 thanks.

yashk2000:
@kk63852 at the place where you've used the toast, replace the toast with a proper snackbar created by you there instead of calling the Snackbarhandler. That way changes made by me will not affect changes in this pull request. Is it ok to do it this way?

kk63852:
@yashk2000 checkout my latest commit.

yashk2000:
LGTM

app/src/main/java/org/fossasia/phimpme/utilities/Utils.java
```diff
@@ -15,11 +15,14 @@
 import android.provider.MediaStore;
 import android.speech.RecognizerIntent;
 import android.support.annotation.NonNull;
+import android.support.design.widget.Snackbar;
 import android.util.Base64;
 import android.view.View;
 import android.webkit.MimeTypeMap;
+import android.widget.Toast;
```
yashk2000: I think this import will be unused now. If it is unused remove it.

abishekvashok:
Thanks

PR #2482, opened 2019-02-03
accent color change dialog button changed to yes and no

### LINKED ISSUE #2477: accent color dialog issue.
**Actual Behaviour**

after setting up the primary color from settings a dialog pops up asking the user to set the accent color same as primary color. the dialog buttons currently displayed are "cancel" and "ok".

**Expected Behaviour**

the dialog buttons should be "yes" and "no" because the user is given a choice here.

**Steps to reproduce it**

open settings and set up primary color.  after this the dialog would pop up and thedialog buttons displyed are cancel and no.

**Screenshots of the issue**

whatsapp image 2019-02-02 at 17 27 48

yes.

kk63852:
Fixed #2477 

Changes: 
accent color change dialog button changed to yes and no
Screenshots of the change: 
whatsapp image 2019-02-04 at 03 21 34


abishekvashok:
Thanks

PR #2486, opened 2019-02-04
Fix for issue#2484: Removed hardcoded strings from xml files.

### LINKED ISSUE #2484: Hardcoded strings in xml files.
**Actual Behaviour**

There are hardcoded strings in some xml files.

**Expected Behaviour**

These strings should be called from string resources.

YES.

yashk2000:
Fixed #2484 
Changes: Hardcoded strings in xml files have with replaced with string resource values.


abishekvashok:
Thanks

PR #2489, opened 2019-02-05
deleting from favourite now requires selecting an image first

### LINKED ISSUE #2485: favourite photos remove from favourites option issue
**Actual Behaviour**

when remove from favourites option is executed, it removes all the media from the favourite folder even when no media is selected.

**Expected Behaviour**

only when some media is selected then this option should be executed.

**Steps to reproduce it**

add some media to favourite folder. navigate to favourite photos folder and tap on delete option when no media is selected from the folder. all the images would be removed from favourite without selecting them.

yes.

kk63852:
Fixed #2485 

Changes:
delete action made invisible unless some media is selected.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -1887,6 +1887,9 @@ public class LFMainActivity extends SharedMediaActivity {
         menu.findItem(R.id.zipAlbumButton).setVisible(editMode && !all_photos && albumsMode && !fav_photos && !hidden &&
                 getAlbums().getSelectedCount() == 1);
         menu.findItem(R.id.delete_action).setVisible((!albumsMode || editMode) && (!all_photos || editMode));
+        if (fav_photos){
+            menu.findItem(R.id.delete_action).setVisible(editMode);
+        }
         if(fav_photos && favouriteslist.size() == 0 ){
             menu.findItem(R.id.delete_action).setVisible(false);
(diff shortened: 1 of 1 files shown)


angmas1:
When no images were selected, that button removed all images from favorites which I think served a purpose but wasn't clear for the user. Either you should change the dialog text when delete button is pressed to ask for removing all the photos from favorites or you should make the delete button disappear when no photos are selected.

angmas1:
Okay, then why not make delete button disappear when no photos are selected i.e. when not in editmode

kk63852:
okay i'll make the changes.

kk63852:
i tried making it disappear but to make it disappear it will need a lot of unnecessary code change because favorite album has no separate code, it shares its menu item with album mode.
the message could be changed but the response would be the same that is it would empty the folder if no media is selected but upon selecting some media is it will be deleted by performing this action. therefore i think its better this way only. @angmas1

angmas1:
Okay, no problem.

abishekvashok:
> i tried making it disappear but to make it disappear it will need a lot of unnecessary code change because favorite album has no separate code, it shares its menu item with album mode.
the message could be changed but the response would be the same that is it would empty the folder if no media is selected but upon selecting some media is it will be deleted by performing this action. therefore i think its better this way only. @angmas1

Then we could just remove this PR and the issue. Bcz, I expect the trash button to remove all favourites. Displaying the button and displaying an alert saying that you need to select before you click, is not a good UI.

kk63852:
@abishekvashok i made the delete button invisible unless some media is selected.

PR #2499, opened 2019-02-07
removed password from favourite and upload history

### LINKED ISSUE #2498: removing media from favourite folder and upload history while security is on requires password.
**Actual Behaviour**

removing media from favourite folder and upload history while security is on requires password.

**Expected Behaviour**

this should not require password as only a copy of image is being deleted.

**Steps to reproduce it**

add some image to favourite folder while security is on. tap on delete. it would require password yo proceed.

yes.

kk63852:
Fixed #2498 

Changes: 
password feature removed from favorite and upload history while deleting some media.


kk63852:
@sauravvishal8797 i removed password from favourites folder only.

PR #2514, opened 2019-02-10
Fix for issue#2513: Made a convention common for all security dialogs. 

### LINKED ISSUE #2513: Insert password dialog issue.
**Actual Behaviour**

The ok and cancel buttons in insert password dialog are placed opposite of where they are in other dialogs. 

**Expected Behaviour**

In most dialogs, ok is to the left and cancel to the right. Insert password dialog should follow the same convention.

YES. I will also check other dialogs with ok cancel or yes no buttons and make all of them follow a common convention.

yashk2000:
Fixed #2513 
Changes: A common convention has been made for all security related dialogs. The ok button is on the left and cancel on the right.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -682,8 +682,8 @@ public class LFMainActivity extends SharedMediaActivity {
                             }
                         }
                     });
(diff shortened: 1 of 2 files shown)


sauravvishal8797:
@yashk2000 Is this necessary?

yashk2000:
@sauravvishal8797 I think it should be done as in some places, in similar dialogs, the position of ok and cancel buttons were opposite. The user tends to get accustomed to a particular convention. Hence having to get used to two different conventions for similar dialogs would not be good.

abishekvashok:
Thanks

PR #2531, opened 2019-02-11
Fix #2210 Added feature to remove multiple photos from favourites from album itself.

### LINKED ISSUE #2210: Photo cannot be marked as unfavourite from the album itself
**Actual Behaviour**

Photo can be marked as unfavourite only from favourite option in navigation drawer.

**Expected Behaviour**

User should have option to mark that photo as unfavourite from that album itslef instead of going to navigation drawer option.

Yes , i am working on this issue.

angmas1:
Fixed #2210 

Changes: LFMainActivity.java, menu_albums.xml, strings.xml Added option to remove selected photos from favorites in options menu. This option is visible only when any one of the selected photos is in favorites.

Screenshots of the change: 
51489028-1e939380-1dcd-11e9-9cc7-888d9d39bde5

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -781,7 +781,6 @@ public class LFMainActivity extends SharedMediaActivity {
         super.onCreate(savedInstanceState);
         Log.e("TAG", "lfmain");
         ButterKnife.bind(this);
-
         navigationView = findViewById(R.id.bottombar);
         favicon = findViewById(R.id.Drawer_favourite_Icon);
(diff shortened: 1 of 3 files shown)


yashk2000:
@angmas1 I do not think it should be implemented in this way. If the option has to be added to the menu, it should be shown only if all the selected images belong to favorites. It will not be appropriate if the user selects several images and only 1-2 of them are favorites and rest are not but still remove from favorites shows up in the meu.

angmas1:
So you want me to make it such that the option to remove from favorites shows only when all selected photos are in favorites?

yashk2000:
@angmas1 yes, that's what I meant.

angmas1:
@yashk2000 made the changes you suggested.

PR #2546, opened 2019-02-13
Fix #2525 Back button get stuck when there is deny of permission

### LINKED ISSUE #2525: Back button get stuck when there is deny of permission
**Actual Behaviour**

Back button is not working when deny the camera permission (click on Never ask again).

**Expected Behaviour**

Back button should work.

**Steps to reproduce it**

Step 1. App open first time.

Step 2. Click on camera tab it will ask for permission

Step 3. Deny the permission and also click on never ask again after that back button is not working,

Yes.

sparrow007:
Fixed #2525 

Changes: [calling the runtime permission in onResume can create a infinite loop if user deny the permission so i check that if user deny permission temporary or permanent and return if permanent deny the permission.]

GIF of the change:

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/opencamera/Camera/CameraActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/opencamera/Camera/CameraActivity.java
@@ -2863,7 +2863,6 @@ public class CameraActivity extends ThemedActivity implements AudioListener.Audi
                     if (ActivityCompat.shouldShowRequestPermissionRationale(CameraActivity.this,
                             Manifest.permission.CAMERA)) {
                         // now, user has denied permission (but not permanently!)
-
                     } else {
                         permanentDenyPermission = true;
                     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/opencamera/Camera/CameraActivity.java
```diff
@@ -2855,6 +2860,13 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String permissi
                         Log.d(TAG, "camera permission granted");
                     preview.retryOpenCamera();
                 } else {
+                    if (ActivityCompat.shouldShowRequestPermissionRationale(CameraActivity.this,
+                            Manifest.permission.CAMERA)) {
+                        // now, user has denied permission (but not permanently!)
+
```
yashk2000: Please remove this extra line.

app/src/main/java/org/fossasia/phimpme/opencamera/Camera/CameraActivity.java
```diff
@@ -2855,6 +2860,13 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String permissi
...
+                    } else {
+                        permanentDenyPermission = true;
+                    }
```
yashk2000: There is no need to add braces here, it is a single statement. Just type it in front of `else`.

sparrow007: i got another idea , will push in a minute.

PR #2559, opened 2019-02-14
fix: Local folders to be secured can be chosen again.

### LINKED ISSUE #2368: Choosing local folders to be secured can be done only once.
**Actual Behaviour**

When the user selects secure local folders, the user can just choose the folders to secure once. If the user has to remove/add security to any other local folder, local folders security has to be turned off and set again by selecting all the desired folders.

**Expected Behaviour**

Once security on local folders is on, an option should be provided to the user to add or remove folders which can be secured.

**Steps to reproduce it**

1) Open security setting and set a password if not set already.
2) Activate security for local folder. A menu is shown for choosing folders to be secured.
3) Close the menu after choosing desired folders.
4) Now if security is to be added or removed from a folder, it cannot be done without removing the security on local folders.

Yes.

yashk2000:
Fixed #2368 

Changes: Now local folders can be secured again without switching off security. Also now if security on local folders is turned off, the folders which were previously secured are remembered. When security is switched on later, those folders are already checked.

Screenshots of the change: 

screencap
screencap1
screencap2


theabhishekavi:
@yashk2000 According to my opinion, instead of getting the secured folder list everytime in recyclerview(which is a bad practice) ,we can get the secured folder list just before creating the recyclerview and we can make the albums secured even before inflating the view. So that when recyclerview is created ,the checkbox corresponding to the secured album gets marked automatically. In that way we don't need to call secured folder list everytime in recyclerview.

yashk2000:
@theabhishekavi I agree, it is a better way. I will update my pull request with the changes.

theabhishekavi:
@yashk2000 one more thing that i want to say, we can get the history of previously locked folders in just 3-4 lines of code. All the codes are so well written that we can easily implement new features... So we don't need much changes in code. And instead of creating an option to choose folders we can create an onclicklisteners on secured folders itself which will create the same dialog box. So we dont need an extra folder to be created.

theabhishekavi:
And there is one more fault in that section is that when we directly toggle off the button. Secured folder is disable but in shared preferences the details of previously locked folder is still there which can creates an issue in other activities. So remember to clear the data from sharedpreferneces also if we toggle off the button. Heartily welcome for any more suggestion if you want. :-)

yashk2000:
> And there is one more fault in that section is that when we directly toggle off the button. Secured folder is disable but in shared preferences the details of previously locked folder is still there which can creates an issue in other activities. So remember to clear the data from sharedpreferneces also if we toggle off the button. Heartily welcome for any more suggestion if you want. :-)

I did not clear the data from special preferences deliberately. I think it will be better to remember which folders the user secured the last time. If the user will be setting security again, most probably the user will set security on those folders again in addition to any more folder if needed.

yashk2000:
> @yashk2000 one more thing that i want to say, we can get the history of previously locked folders in just 3-4 lines of code. All the codes are so well written that we can easily implement new features... So we don't need much changes in code. And instead of creating an option to choose folders we can create an onclicklisteners on secured folders itself which will create the same dialog box. So we dont need an extra folder to be created.

I did not do this as the user might just want to turn off security on local folders. It is better to have a separate choose folder textview to allow the user to choose folders again.

theabhishekavi:
yes,you are right don't clear the data from sharedpreferences. I think i was talking about some other activity which is not needed there. You should just get secured folder list from shared preferences and set album.setsecured(true) in case of secured albums. And do this implementation before recyclerview as said earlier.

yashk2000:
Yes. I did it. I'll push it soon.

theabhishekavi:
And can we make securedfolder itself to be clickable if password on folder is on?

yashk2000:
That is a separate issue since even now when the dialog pops up, the securedfolder is not clickable. I'll open a separate issue for it.

theabhishekavi:
So instead of making a new folder can we make securedfolder itself be clickable to open the dialog box list of securedfolder?

yashk2000:
> So instead of making a new folder can we make securedfolder itself be clickable to open the dialog box list of securedfolder?

Can you please elaborate this a bit, I did not understand this?

yashk2000:
Also I made the changes to this pr. Is it fine now?

yashk2000:
Do you mean that we should make "Local Folders" itself clickable if security is on? and show the dialog if it is clicked?

theabhishekavi:
Yes,exactly. And in that way we don't need an extra choose folders.

theabhishekavi:
> Also I made the changes to this pr. Is it fine now?

Now the code looks good to me, but still i think instead of creating a new textview,you should make Local Folder itself be clickable if security is on.

theabhishekavi:
> Now the code looks good to me, but still i think instead of creating a new textview,you should make Local Folder itself be clickable if security is on.

@abishekvashok what do you think?

yashk2000:
> > Also I made the changes to this pr. Is it fine now?
> 
> Now the code looks good to me, but still i think instead of creating a new textview,you should make Local Folder itself be clickable if security is on.

No that shouldn't be done. We should leave it so that user can easily toggle the button on/off. Making a new textview makes it easier for the user to understand what is to be done to secure more folders.

theabhishekavi:
Okay.

yashk2000:
@abishekvashok @sauravvishal8797 please review.

PR #2578, opened 2019-02-16
fix: An alert is shown to the user asking whether app should be restarted or not.

### LINKED ISSUE #2521: App restarts without giving warning when Camera2 API is enabled or disabled.
**Actual Behaviour**

The app restarts suddenly when Camera2 API is enabled or disabled.

**Expected Behaviour**

The user should be shown a dialog asking whether the user wants to restart the app immediately or later.

**Steps to reproduce it**

Go to camera settings. Enable or disable Camera2 API. App restarts suddenly without giving the user a warning.

YES.

yashk2000:
Fixed #2521 

Changes: The user is given the choice of restarting the app immediately or later. If the user does not want to restart the app immediately, the changes are made when the app is closed and restarted by the user whenever the user closes the app and restarts it manually.

Screenshot of the change: 

screencap


abishekvashok:
Thanks

PR #2606, opened 2019-02-21
fix: Made location text color same as accent color.

### LINKED ISSUE #2603: Color of image location in image details.
**Describe the bug**
The color of image location text is always yellow.

**Expected behavior**
Text color of image location should match app theme.

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Nokia 7 plus|
|Android Version|Android 9    |

- [X] Yes
- [ ] No

yashk2000:
Fixed #2603 

Changes: Location text color in image details is now the same as an accent color for the app instead of staying yellow all the time.

Screenshots of the change: 

screencap


kk63852:
when i opened this issue why was i told to leave it as it is and now you have done the same thing?

yashk2000:
> when i opened this issue why was i told to leave it as it is and now you have done the same thing?

You had opened an issue according to which the text color would have been made the same as the other texts in description. Then there would have been no differentiating factor. I opened an issue to make the color the same as the accent color.

PR #2607, opened 2019-02-21
Fix #2595 in slideshow set timer dialog, keyboard enter can be used as ok button

### LINKED ISSUE #2595: After inserting "enter time interval" in slideshow option, Keyboard's "tick mark" key can be used instead of pressing "ok"

After Entering the time in the "enter time interval" field in slideshow option, User needs to press "ok" button, user should be allowed to use keyboard's "tick mark" key instead of pressing "ok" 
 
**Screenshots**

- [x] Yes
- [ ] No
- Other:

neel1998:
Fixes #2595 

Changes: added click listener for keyboard enter

Screenshots of the change:


abishekvashok:
Thanks

PR #2615, opened 2019-02-22
Fix #2612 in slide show, toast replaced by snackbar and keyboard doesn't close on invalid time interval

### LINKED ISSUE #2612: Slideshow dialog, if keyboard tick is pressed, dialog closes even if input is incorrect.
**Describe the bug**
When enter button on the keyboard is pressed, the dialog keyboard disappears no matter what the input is.

**To Reproduce**
Open slideshow, enter an incorrect value. Press enter on the keyboard, the keyboard will disappear.

**Expected behavior**
The keyboard should not disappear. It should stay there for the user to input correct data. And also instead of a toast, a snackbar should be shown when slideshow starts after pressing enter on the keyboard.

- [ ] Yes
- [X] No
- Other: @neel1998 would you like to work on this issue since you fixed a similar one today?

neel1998:
Fixed #2612 

Changes: Toast replaced with snackbar, key listener added on alert Dialog EditText

Screenshots of the change:


neel1998:
@yashk2000  review please

Issue #2593, opened 2019-02-19
Replacing Snackbar.make() in the app with SnackbarHandler.


yashk2000:
**Describe the chore**
The snackbars made in the app directly should be made using SnackbarHandler.

**Would you like to work on the issue?**

- [X] Yes
- [ ] No

PR #2619, opened 2019-02-22
fix: Snackbar is shown if password is set again after removal.

### LINKED ISSUE #2611: No indication that a password has been set.
**Describe the bug**
If a password has been set once, then removed and set again, there is no indication that a password has been set.

**To Reproduce**
Set a password once, remove it, set it again. No indication is shown that password has been set.

**Expected behavior**
An indication should be given to show that the password has been set.

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Nokia 7 plus|
|Android Version|Android 9    |

- [X] Yes
- [ ] No

yashk2000:
Fixed #2611 

Changes: A snackbar is shown if a password is set again after removal.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
@@ -102,11 +102,10 @@ public class SecurityActivity extends ThemedActivity {
                         swActiveSecurity.setChecked(true);
                         SP.putBoolean(getString(R.string.preference_use_password), true);
                         toggleEnabledChild(true);
-                    }
-                    else
-                    setPasswordDialog();
-                }
-                else {
+                        Snackbar.make(findViewById(android.R.id.content), "Password Set", Snackbar.LENGTH_SHORT)
+                                .show();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/SecurityActivity.java
```diff
@@ -102,11 +102,10 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
...
                         toggleEnabledChild(true);
-                    }
-                    else
-                    setPasswordDialog();
-                }
-                else {
+                        Snackbar.make(findViewById(android.R.id.content), "Password Set", Snackbar.LENGTH_SHORT)
+                                .show();
+                    } else setPasswordDialog();
+                } else {
```
neel1998: Cant you use SanckBar Handler here as it is used throughout the code. Also put "password set" string into strings.xml instead of hardcoding

yashk2000: I will make use of snackbar handler when #2627 is merged. I will fix this along with issue #2593.

PR #2627, opened 2019-02-23
fix: Now snackbars are visible.

### LINKED ISSUE #2467: Inactive snackbars in the app.
**Actual Behaviour**

Currently, there are many redundant and inactive snackbars in the app.

**Expected Behaviour**

The redundant snackbars should be removed and the inactive ones should be corrected so as to be displayed when required.

YES.

yashk2000:
Fixed #2467 
Changes: The snackbars which were earlier not being displayed are now being displayed.


abishekvashok:
Referencing a pr? Maybe fix that there itself.

yashk2000:
@abishekvashok both the issues are entirely different.

PR #2684, opened 2019-03-12
Fixes for NDK r19.

DanAlbert:
* Use a supported STL.
* Fix bad style reference (aapt2 catches this).
* Update gradle plugin to one compatible with NDK r16+.

This still isn't sufficient because the OpenCV prebuilts are gnustl
only, but it's a starting place for anyone that can tackle that.

### DIFF
--- a/app/src/main/jni/Application.mk
+++ b/app/src/main/jni/Application.mk
@@ -1,5 +1,5 @@
 APP_OPTIM := release
 APP_ABI := armeabi-v7a
-APP_STL := gnustl_static
+APP_STL := c++_static
 APP_CPPFLAGS := -frtti -fexceptions
 APP_PLATFORM := android-25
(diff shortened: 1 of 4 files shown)


DanAlbert:
The second and third fix here allow this project to be built with NDK r16/r17, but the getting the project building with r18+ is more involved because the OpenCV prebuilts you link to in the readme are built against gnustl rather than libc++, and gnustl support is gone as of r18.

I'm not sure if you can update to a newer release of those libraries or if you'll need to build them yourselves, but if you want the project to build with NDK r18+ something will need to be done. If you don't, let me know and I'll split that part of this change out so you can at least use r16/r17.

yashk2000:
@abishekvashok these changes are not enough to update the build gradle. I tried doing it this way, but a lot of other changes have to made. We need to update other libraries as well. And remove some redundant codes which are in the gradle build version being used currently but are not needed in the latest version.

mariobehling:
Ok, thank you. Merging this as WIP. @yashk2000 Could you please help to take of the outstanding items?

PR #2686, opened 2019-03-16
removed unnecessary else if statements and the strings related to it.

### LINKED ISSUE #2685: unnecessary else if statements in set slideshow dialog.
**Describe the bug**

unnecessary else if statements in set slideshow dialog.
**To Reproduce**
open the SingleMediaActivity and navigate to SetSlideshowDialog. there are some unnecessary else if statements in the dialog.

**Expected behavior**

these statements are redundant, hence these should be removed.

- [X] Yes
- [ ] No
- Other:

kk63852:
Fixed #2685 

Changes:
removed unnecessary else if statements and the strings related to it.


abishekvashok:
Why are we removing this?

kk63852:
We are removing these statements because  these statements are not needed anymore as we get error message just by entering wrong input to slideshow dialog.

PR #2705, opened 2019-03-31
fix: Fixed the position of description.

### LINKED ISSUE #2528: Description of image no visible in landscape mode.
**Actual Behaviour**

In landscape mode, the description of the image is not visible in the details fragment.

**Expected Behaviour**

Description of the image should be visible in details fragment even in landscape mode.

YES.

yashk2000:
Fixed #2528 

Changes: The complete padding on the linear layout which was causing visibility issues has been removed. Instead, padding was added individually to the views which required it.

Screenshots of the change: 

image


abishekvashok:
Thanks

PR #2708, opened 2019-04-01
fix: A photo cannot be renamed with the same name.

### LINKED ISSUE #2610: Renaming an image with the same name gives no notification.
**Describe the bug**
While renaming an image, if the image is renamed with the same name, the image is still being renamed.

**To Reproduce**
Try to rename an image with the same name as it has. No notification is given.

**Expected behavior**
The image should not be renamed and a snack bar should be shown saying that there was no change in the name of the image.

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Nokia 7 PLus|
|Android Version|Android 9   |

- [X] Yes
- [ ] No

yashk2000:
Fixed #2610 

Changes: Now a photo cannot be renamed if the user renames with the same name.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
@@ -1176,7 +1176,7 @@ public class SingleMediaActivity extends SharedMediaActivity implements ImageAda
                 final File file = new File(currentpath);
                 int indexofdot = file.getPath().lastIndexOf(".");
                 int indert = file.getPath().lastIndexOf("/");
-                String namefile = file.getPath().substring(indert + 1, indexofdot);
+                final String namefile = file.getPath().substring(indert + 1, indexofdot);
                 final String imageextension = file.getPath().substring(indexofdot + 1);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/phimpme/gallery/activities/SingleMediaActivity.java
```diff
@@ -1281,28 +1281,33 @@ public void onClick(DialogInterface dialogInterface, int i) {
                                 alertDialog.show();
                                 AlertDialogsHelper.setButtonTextColor(new int[]{DialogInterface.BUTTON_POSITIVE, DialogInterface.BUTTON_NEGATIVE, DialogInterface.BUTTON_NEUTRAL}, getAccentColor(), alertDialog);
                             } else {
-                                int index = file.getPath().lastIndexOf("/");
-                                String path = file.getPath().substring(0, index);
```
kk63852: why were these lines removed only to be added later?

yashk2000: I used an additional if condition. So all the lines have been shifted down. It's just appearing to be this way here. I have just added lines, not deleted any.

kk63852: cool.

PR #2719, opened 2019-04-05
fix: Added a convert to PDF option.

### LINKED ISSUE #2718: Convert to PDF option.
Currently, there is a print option using which images can be converted into pdfs, but the user will not know the functionality as there is no direct option to convert an image into a PDF. There should be an option called `Convert to PDF` so that user can convert images to pdfs.

- [X] Yes
- [ ] No

yashk2000:
Fixed #2718 

Changes: An option has been added which uses the same function as printing but directly tells the user that the feature can be used to convert an image to pdf and save the image as pdf.

Screenshots of the change: 

screencap


abishekvashok:
Thanks!

PR #2724, opened 2019-04-18
Fix #2721 disable swipe to refresh

### LINKED ISSUE #2721: Disable swipe to Refresh when item is selected in favourite section
**Describe the bug**

Issue #1277 has full description where swipe to refresh should disable while selecting the items but here it should implement in favorite section.

- [x] Yes
- [ ] No
- Other:

sparrow007:
Fixed #2721

Changes: [Add condition so it can disable the swipe to refresh layout in favorite section as well all photos ]

GIF of the change:


abishekvashok:
Thanks

PR #2737, opened 2019-05-17
fix: Removed security settings feature.

### LINKED ISSUE #2736: Delete Password Feature and give Access to Everything Without Password
People already use a password etc. to access their phone. We only have a small team and should not take care of features that are already taken care of by the system.

Please take out the password feature entirely.

Other reasons:
* it is not the task of the app to control access within the device (anyone who has access to the phone can easily circumvent that anyways)
* we don't have an easy way to reset passwords through email reset (like online services)
-> We should not take our time to maintain password features and it is not a high requirement stopping users from flocking to the app.

yashk2000:
Fixed #2736 

Changes: Security features have been removed from the app completely.


yashk2000:
@iamareebjamal @mariobehling please review.

PR #2760, opened 2019-06-05
chore: Update build gradle version to 3.4.1 and updated to OpenCV4.0.1

### LINKED ISSUE #2744: App contains only 32 bit native binaries.
**Describe the chore**
The app currently contains only 32-bit native binaries. These will be banned after August. They should be replaced with 64-bit binaries.

- [X] Yes
- [ ] No

yashk2000:
Fixed #2744 

Changes: Updated build gradle version beeing and openCV version to 4.0.1 to include 64-bit native libraries.


yashk2000:
@iamareebjamal @abishekvashok @sauravvishal8797 please review.

iamareebjamal:
LGTM, but why 4.0.1

Since this is a big PR, why not upgrade to latest 4.1.0 instead of creating another big PR for this

Secondly, can't the files in `include` directory omitted and copied from the opencv sdk?

I suppose you have copied them from it, right? Why keep code in the repo that we don't maintain and have to update on each library update?

yashk2000:
@iamareebjamal yes, we can omit the `include` directory.

I tried with 4.1.0 after 4.0.1 earlier, but it gave me some errors which were not faced with 4.0.1. So I sent the pr with 4.0.1. I'll try 4.1.0 once again.

yashk2000:
And also omit the `include` directory.

yashk2000:
@iamareebjamal omitting the `include` directory will be done by adding `/app/src/main/jni/include/` to the .gitignore file right?

iamareebjamal:
Your PR should have about 10 files changed once you have omitted the include directory.

yashk2000:
@iamareebjamal yes sir. I added `/app/src/main/jni/include/` to the `.gitignore` file, but the folder is still here.

iamareebjamal:
It obviously will be. You have to delete it. Also, don't squash and force push because if it doesn't work, you won't be able to revert

iamareebjamal:
Also, don't delete the folder in this PR, just revert the changes

yashk2000:
> It obviously will be. You have to delete it. Also, don't squash and force push because if it doesn't work, you won't be able to revert

@iamareebjamal sir but `/app/src/main/3rdparty/` and `/app/src/main/jniLibs` are also in the `.gitignore`. I do not have to delete these folders before pushing changes.

iamareebjamal:
Because they were never added in git

Once tracked, always tracked, doesn't matter if you add them in .gitignore after. Read about git tracked files and gitignore

yashk2000:
Ohk sir. I'll delete the folder and push the changes. Thanks :)

iamareebjamal:
> Also, don't delete the folder in this PR, just revert the changes

iamareebjamal:
Don't delete. The point is to not have 630 changes in this PR. It should only have changes in makefiles and download scripts

yashk2000:
@iamareebjamal  Sir but without deleting how will we make ignore the `include` file while pushing?
Should that be done in another pr?
Which changes do I revert back to then? Just updating the build gradle version?

iamareebjamal:
> Sir but without deleting how will we make ignore the include file while pushing?
Should that be done in another pr?

Yes

> Which changes do I revert back to then?

All changes should be there except any change in include folder. The opencv script must be updated to copy required files to the include folder

yashk2000:
> The opencv script must be updated to copy required files to the include folder

If I do this, then this error will happen:
`mv: cannot move ‘/home/travis/openCV/opencv/sdk/native/jni/include/opencv2’ to ‘app/src/main/jni/include/opencv2’: Directory not empty
`
For now, until the `include` folder is deleted, should I modify the script to first delete the existing contents of the folder and then move in the new folders?

iamareebjamal:
Still 400+ changes. Same can be done with ` openCVLibrary401/` folder. Please don't add anything we are copying from the opencv SDK

In future, we'll use open cv aar like https://github.com/piruin/opencv-android

So that we don't need this utterly useless step and redundant code in the repo and can manage the dependency in sane way like other "normal" android libraries

```
implementation 'org.opencv:opencv-android:3.4.1'
```

Not doing it currently because the latest version is 3.4.1

yashk2000:
@iamareebjamal openCVLibrary401 is not copied and pasted. It has to imported as a separate module. I think we should leave this because some changes have to made in the manifests and gradle files of both app and openCV module. This might not be very friendly for first-time contributors.

yashk2000:
> In future, we'll use open cv aar like https://github.com/piruin/opencv-android

Yes sir, this is a very good idea.

yashk2000:
@iamareebjamal is it alright now?

yashk2000:
@iamareebjamal sir, is this alright now?

yashk2000:
@iamareebjamal sir, is it ok now?

iamareebjamal:
> openCVLibrary401 is not copied and pasted. It has to imported as a separate module. I think we should leave this

OK, but rename this to `openCV`, so that we don't have to change this every version. I have no idea why it was thought of as a good idea to add version code in module

yashk2000:
Ok sir, on it.

yashk2000:
> OK, but rename this to `openCV`, so that we don't have to change this every version. I have no idea why it was thought of as a good idea to add version code in module

@iamareebjamal done.

PR #2769, opened 2019-06-10
fix: Added snackbars when image is deleted from SingleMediaActivity.

### LINKED ISSUE #2590: snackbar should be displayed if image is deleted permanently
**Describe the bug**

no message is displayed if image is deleted permanently.
**To Reproduce**

1. select few images.
2. Click on 'delete'
3. uncheck the checkbox which ask the item to move it to trashbin.
4. no message is displayed if the images are deleted permanently.

**Expected behavior**

a snackbar should be displayed.

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Samsung J7 next|
|Android Version|Oreo 8.1    |

- [X] Yes
- [ ] No

yashk2000:
Fixed #2590 

Changes: If an image is deleted or moved to trashbin from the SingleMediaActivity, a snackbar is shown. Removed the `showsnackbar` method as we need different messages when albums and images are deleted.

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -1600,24 +1600,6 @@ public class LFMainActivity extends SharedMediaActivity {
     }
   }
 
-  private void showsnackbar(Boolean result) {
-    if (result) {
-      Snackbar snackbar =
-          SnackBarHandler.show(
-              mDrawerLayout,
-              getApplicationContext().getString(R.string.photo_deleted_msg),
-              navigationView.getHeight());
-      snackbar.show();
-    } else {
-      Snackbar snackbar =
-          SnackBarHandler.show(
-              mDrawerLayout,
-              getApplicationContext().getString(R.string.photo_deletion_failed),
-              navigationView.getHeight());
-      snackbar.show();
-    }
-  }
-
   private void checkNoSearchResults(String result, int length) {
     if (length > 0) {
       textView.setText(getString(R.string.null_search_result) + " " + '"' + result + '"');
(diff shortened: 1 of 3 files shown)


yashk2000:
@adityastic @iamareebjamal @abishekvashok @mariobehling please review.

yashk2000:
@iamareebjamal @abishekvashok @mariobehling please review.

mariobehling:
Please provide a screenshot and test apk.

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -2190,22 +2190,26 @@ protected void onPostExecute(Boolean result) {
                 getAlbums().clearSelectedAlbums();
                 mDrawerLayout.setDrawerLockMode(DrawerLayout.LOCK_MODE_UNLOCKED);
                 albumsAdapter.notifyDataSetChanged();
+                Snackbar snackbar =
+                    SnackBarHandler.show(
+                        mDrawerLayout,
```
abishekvashok: You could make it a function and pass the string as a parameter

yashk2000: SnackbarHandler is the function created for this purpose.

yashk2000:
@mariobehling screenshot: All snackbars made are of the same type, for albums, different message.

screencap

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
abishekvashok: Reuse the statement pls, it's being used many times.

yashk2000: But that is how snackbars are created throughout the app. In some cases, the SnackbarHandler functions are used directly, otherwise a snackbar is returned from the class which is stored and then displayed. It has to be this way in order to adjust for the layout and height of the bottom bar.

If there is any other way, could you please suggest?

PR #2777, opened 2019-06-13
fix: Fix travis

yashk2000:
Changes: Attempt to fix the error in the travis script due to which the build is getter errored.


yashk2000:
@iamareebjamal @abishekvashok please advise.

PR #2782, opened 2019-06-19
fix: If photo is not being saved, user is guided to the permission setting to give proper permission.

### LINKED ISSUE #2771: Cannot Save Photo after Editing
User Report: 

I like app appreciate you writing it , but on my Motorola phone G5 plus S , there is no workable option to save the photo after having edited it. when I am in the editing screen , and click the check mark it goes back to the editing screen . When I click the top right icon , nothing happens. when I back out of the app , it says I am leaving without saving and offers to save . I can click to save , as far as I can tell nothing at all happens. If I try to back out again after that it brings up the same option to save or discard dialogue. I can't find information on this on the wiki page .
 Thanks for any thoughts you have

yashk2000:
Fixed #2771 

Changes: If the user is not able to save images, the user is directed to the permission settings.

Screenshots of the change: 

screencap

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/editor/EditImageActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/editor/EditImageActivity.java
@@ -6,8 +6,10 @@ import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.graphics.PorterDuff;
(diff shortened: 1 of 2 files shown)


yashk2000:
@adityastic @iamareebjamal @abishekvashok @mariobehling please review.

PR #2800, opened 2019-07-02
fix: Added a background dialog for delete, copy/move actions

### LINKED ISSUE #2455: progress dialog for many functions like delete, copy.
**Actual Behaviour**

while trying to move or delete large number of files(in multiples of tens) only refreshing is enabled. 

**Expected Behaviour**

there should be a progress dialog to inform the user about the progress of the task being done.

**Steps to reproduce it**

open a album and delete large number of files. refreshing is enabled only.

yes.

yashk2000:
Fixed #2455

Changes: Added a background progress dialog for copy/move and delete actions.

Screenshots of the change: 

screencap

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
@@ -8,6 +8,8 @@ import static org.fossasia.phimpme.utilities.ActivitySwitchHelper.context;
 
 import android.animation.Animator;
 import android.annotation.TargetApi;
+import android.app.Dialog;
+import android.app.ProgressDialog;
 import android.content.ContentResolver;
 import android.content.ContentUris;
 import android.content.Context;
@@ -2042,11 +2044,21 @@ public class LFMainActivity extends SharedMediaActivity {
 
           private boolean succ = false;
           private int imagesUnfav = 0;
+          private Dialog dialog;
 
           @Override
           protected void onPreExecute() {
+            dialog = getLoadingDialog(context, "Deleting", false);
+            dialog.show();
             swipeRefreshLayout.setRefreshing(true);
             super.onPreExecute();
+          }
+
+          private Dialog getLoadingDialog(Context context, String title, boolean canCancel) {
+            ProgressDialog dialog = new ProgressDialog(context);
+            dialog.setCancelable(canCancel);
+            dialog.setMessage(title);
+            return dialog;
           }
 
           @Override
@@ -2190,6 +2202,7 @@ public class LFMainActivity extends SharedMediaActivity {
 
           @Override
           protected void onPostExecute(Boolean result) {
+            dialog.dismiss();
             if (result) {
               // in albumsMode, the selected albums have been deleted.
               if (albumsMode) {
@@ -4503,6 +4516,7 @@ public class LFMainActivity extends SharedMediaActivity {
     // private Snackbar snackbar;
     private ArrayList<Media> temp;
     private Boolean moveAction, copyAction, success;
+    private Dialog dialog;
 
     CopyPhotos(String path, Boolean moveAction, Boolean copyAction, LFMainActivity reference) {
       this.path = path;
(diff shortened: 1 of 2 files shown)


yashk2000:
@adityastic @iamareebjamal @abishekvashok @mariobehling  please review.

app/src/main/java/org/fossasia/phimpme/gallery/activities/LFMainActivity.java
```diff
@@ -2042,13 +2044,23 @@ public void onClick(DialogInterface dialog, int which) {
...
           @Override
           protected void onPreExecute() {
+            dialog = getLoadingDialog(context, "Deleting", false);
+            dialog.show();
             swipeRefreshLayout.setRefreshing(true);
             super.onPreExecute();
           }
 
+          private Dialog getLoadingDialog(Context context, String title, boolean canCancel) {
+            ProgressDialog dialog = new ProgressDialog(context);
```
adityastic: ProgressDialog was depreciated in api 26 as i remember. Use progress bar instead

yashk2000: Is it necessary? ProgressDialog is still working.

iamareebjamal: Deprecated doesn't mean disabled

yashk2000: @iamareebjamal I know. I think that we shouldn't shift to Progress bar because everywhere we are using ProgressDialog. If we shift this to progress bar, we should shift others too to progress bars.

adityastic: > A program element annotated @Deprecated is one that programmers are discouraged from using, typically because it is dangerous, or because a better alternative exists.


It's upto you, but it is suggested as it is more robust to api changes

iamareebjamal: Yeah, not everytime you can have a UI component which lets user dismiss it. See the stackoverflow thread that not everyone is happy with this change,

yashk2000:
@abishekvashok is that really needed? This dialog pops for a very brief amount of time, mostly just a second. The only time you can actually see it properly is while copying photos, that too for just a second or so. 
Do we need to make a progress bar notification for such a short period? The user won't even see it.

yashk2000:
> @abishekvashok is that really needed? This dialog pops for a very brief amount of time, mostly just a second. The only time you can actually see it properly is while copying photos, that too for just a second or so.
> Do we need to make a progress bar notification for such a short period? The user won't even see it.

@abishekvashok what do you think?

abishekvashok:
> @abishekvashok is that really needed? This dialog pops for a very brief amount of time, mostly just a second. The only time you can actually see it properly is while copying photos, that too for just a second or so.
Do we need to make a progress bar notification for such a short period? The user won't even see it.

@yashk2000 it can serve as for further actions in the future, so I guess it would be a better investment on our part. Let's do it the correct way instead of the easy way

yashk2000:
@abishekvashok I was not doing the easy thing, just following the current convention in the app.
I still did not get the purpose of using the notification method though. I have tested the places where this progress dialog pops up, and everywhere, it comes for a very small period of time, and most often, it isn't even visible. 
Could you please tell me the need for the notification progress bar in such cases where it won't even be visible?

abishekvashok:
> I still did not get the purpose of using the notification method though. I have tested the places where this progress dialog pops up, and everywhere, it comes for a very small period of time, and most often, it isn't even visible.

It's about providing a better UX, a better standard.

yashk2000:
@abishekvashok ok. I'll make a notification dialog bar for these changes.

yashk2000:
@abishekvashok @iamareebjamal I tried a few methods to display the notification progress bar. The app built successfully, but when I run the app, and the copy operation, I am unable to see the notification progress bar.

iamareebjamal:
I don't know if we can invest that much time in notification updates. This is not that long a process. Also, we are not handling progress updates and the ability to cancel tasks, for which notification status are used. Without these capabilities, notifications are as useful as a progress dialog. Once we are that much stable and can handle canceling and progress, we can show notifications. Also, AFAIK, we have yet to handle what would happen if a user opens a file that has since been deleted or copied as the notification status does not block the UI.

YAGNI - You ain't gonna need it

Think twice before adding a feature, not just because you think you're going to need it

Either we have to handle everything before adding this feature. Which will delay this PR by a week at least, or we do a simple thing and move to other more important bugs and features and come back when we have that much time and effort on us to invest in this requirement

yashk2000:
@iamareebjamal so I shouldn't implement that feature right now. I guess then the progress dialogs being shown right now are good enough. If it's fine, can this pr be merged then?

Issue #1181, opened 2017-09-20
Launching app on F-Droid


codedsun:
**Actual Behaviour**
The app has to be launched on F-Droid

**Expected Behaviour**

**Screenshots of the issue**

**Would you like to work on the issue?**

Yes ! @anantprsd5

codedsun:
@anantprsd5 @mohitmanuja @pa1pal  This has to be done ! I will be needing help as @pa1pal is offline since days.

Kristinita:
+1

abishekvashok:
@codedsun any help i can deliver?

codedsun:
@abishekvashok Yes, I had a vagrant error in the linux, I was involved in this process.I contacted their IRC Channel etc but no help was recieved. It would be great if we could take this issue together and solve this

abishekvashok:
What priority software do we use? We don't right?

realmain:
[Here](https://gitlab.com/fdroid/rfp) is where you submit a package for F-Droid btw.

akavel:
"Closed solved"? Or "closed wontfix"? Or what? Is it there on F-Droid yet? [I don't seem to see it on the list](https://f-droid.org/en/packages/33/)

realmain:
He closed this and opened it under #2018 for some reason

sauravvishal8797:
@realmain Actually we had 2-3 issues regarding this matter, so I thought of opening a fresh one mentioning all the requirements for opening the merge request for Fdroid inclusion. Nevertheless, I'd close the #2018 and reopen this. And regarding the Fdroid inclusion, it is certainly at the top of our priority list. And at the moment we need to make some changes to our master branch and we're working on it, and once the changes are successfully implemented we're good to go and I'll be opening a merge request to Fdroid's repo for inclusion. Hopefully in a day or two we'd resolve the master branch issue and open the merge request.

moreystraus:
If you release a new apk you could list it on the [izzy](
https://apt.izzysoft.de/fdroid/index.php) repo straight away.

Robert321245:
Guys
Keep in mind that publish the app in the F-Droid store would give lot of visibility to the app in the Free and Open Source community and, in consequence, would bring new developers to the project.

mariobehling:
@Robert321245 Could you help to make it happen bringing this app to Fdroid, please?

iamareebjamal:
Requirements for F-Droid:

[...]

PR #2833, opened 2019-07-20
chore: Add fdroid product flavor

yashk2000:
- First attempt to add flavor to make an fdroid apk

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,6 +77,18 @@ android {
         }
     }
 
+    flavorDimensions "default"
+
+    productFlavors {
+        fdroid {
+            dimension "default"
+        }
+
+        playStore {
+            dimension "default"
+        }
+    }
+
     packagingOptions {
         exclude 'META-INF/maven/org.apache.maven/maven-artifact/pom.xml'
         exclude 'licenses/javolution.license.TXT'


iamareebjamal:
This is no-op. Move the google dependencies to playStore and OSS dependencies to fdroid

yashk2000:
@iamareebjamal the only google dependency  we are using is exoplayer right?
Rest of the dependencies are open source. And according to [this](https://f-droid.org/en/docs/FAQ_-_App_Developers/#which-libraries-and-dependencies-are-good-to-use) Android support repositories are allowed.

iamareebjamal:
exoplayer is open source

iamareebjamal:
Where are we using google-services.json? If we're not using it, remove it

yashk2000:
@iamareebjamal I checked the dependencies. I did not find the declaration for google services there. Should I remove the file then?

yashk2000:
@iamareebjamal https://github.com/fossasia/phimpme-android/pull/2835

iamareebjamal:
Then we are not using any proprietary library. Let's close this and prepare adding to F-Droid. You'll have to see what other apps using OpenCV do and add a download and moving step in between

yashk2000:
@iamareebjamal we cannot add the app directly to fdroid? OpenCV is open source.

iamareebjamal:
```
git clone https://github.com/fossasia/phimpme-android
cd phimpme-android
./gradlew assembleRelease
```

Run these commands and tell me the output. This is what fdroid will run. If this runs without any error, then yes, you can add the app directly in fdroid

yashk2000:
On it.

yashk2000:
@iamareebjamal we need to add the other files, `jniLibs`, `staticlibs`, `3rdparty` and `jni/include` too.

iamareebjamal:
> You'll have to see what other apps using OpenCV do and add a download and moving step in between

Exactly what I said before

yashk2000:
There used to be an app called OpenCV manager, but it is no longer there. I'll hunt for more ways to use OpenCV today afternoon. 
I'm sorry for the delay because of my college periodical exams this week. I'll compensate for it in the coming week.

iamareebjamal:
This dicussion should be in #1181

PR #2855, opened 2019-08-13
fix: Replace snackbars with snackbarshandler

### LINKED ISSUE #2593: Replacing Snackbar.make() in the app with SnackbarHandler.

yashk2000:
Fixed #2593 

Changes: Replace snackbars with snackbarshandler

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
@@ -187,8 +187,7 @@ public class AccountActivity extends ThemedActivity
 
   @Override
   public void showError() {
-    Snackbar snackbar = SnackBarHandler.show(coordinatorLayout, getString(no_account_signed_in));
-    snackbar.show();
+    SnackBarHandler.create(coordinatorLayout, getString(no_account_signed_in)).show();
   }
 
   @Override
@@ -251,9 +250,9 @@ public class AccountActivity extends ThemedActivity
 
         case DROPBOX:
           if (CLOUDRAIL_APP_KEY == null || CLOUDRAIL_APP_KEY.equals("")) {
-            Snackbar.make(
+            SnackBarHandler.create(
                     findViewById(android.R.id.content),
-                    R.string.Cloudrail_License_key,
+                    getString(R.string.Cloudrail_License_key),
                     Snackbar.LENGTH_SHORT)
                 .show();
           } else signInDropbox();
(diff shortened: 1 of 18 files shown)


yashk2000:
@adityastic please review.

iamareebjamal:
Why?

yashk2000:
We have a separate class which we use for handling snackbars across the app, SnackbarHandler. Some of the snackbars being used currently were just made using snackbar.make, instead of the Handler class for snackbars. So I converted those to SnackbarHandler.

iamareebjamal:
What advantage does SnackbarHandler have?

yashk2000:
@iamareebjamal it was created to have a common class to create all kinds of snackbars except those with actions. We use it to create snackbars with different bottom margins. So all these functions we created in this common class, `SnackbarHandler`, and it is used in almost all places in the app, except the ones which I fixed in this pr.

app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -251,11 +251,12 @@ public void onItemClick(final View childView, final int position) {
...
-            Snackbar.make(
+            Snackbar snackbar =
+                SnackBarHandler.show(
                     findViewById(android.R.id.content),
-                    R.string.Cloudrail_License_key,
-                    Snackbar.LENGTH_SHORT)
-                .show();
+                    getString(R.string.Cloudrail_License_key),
+                    Snackbar.LENGTH_SHORT);
+            snackbar.show();
```
iamareebjamal: This is ridiculous. Why calling show 2 times. It should only be called once

yashk2000: The earlier `.show` is `SnackBarHandler.show()`. This creates a snackbar and returns it. The to show it, we use the second `.show()`.

iamareebjamal: Then it should not be called show. It should be create or make

yashk2000: Ok. I'll do that and update the pr.

yashk2000: Done.

app/src/main/java/org/fossasia/phimpme/accounts/AccountActivity.java
```diff
@@ -187,7 +187,7 @@ public void setUpAdapter(@NotNull RealmQuery<AccountDatabase> accountDetails) {
 
   @Override
   public void showError() {
-    Snackbar snackbar = SnackBarHandler.show(coordinatorLayout, getString(no_account_signed_in));
+    Snackbar snackbar = SnackBarHandler.create(coordinatorLayout, getString(no_account_signed_in));
     snackbar.show();
```
iamareebjamal: No need to save in a variable. Directly call show

yashk2000:
@iamareebjamal please merge.

PR #2896, opened 2019-10-19
feat: Removed volley and shifted api calls to retrofit

### LINKED ISSUE #2895: Move to Retrofit
Remove volley and move to retrofit

- [x] Yes
- [ ] No
- Other:

codedsun:
Fixed #2895 
Changes: Removes volley and shifts to retrofit

Screenshots of the change:

### DIFF
--- a/app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
+++ b/app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
@@ -54,14 +54,6 @@ import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import butterknife.BindView;
 import butterknife.ButterKnife;
-import com.android.volley.AuthFailureError;
-import com.android.volley.DefaultRetryPolicy;
-import com.android.volley.Request;
-import com.android.volley.RequestQueue;
-import com.android.volley.Response;
-import com.android.volley.VolleyError;
-import com.android.volley.toolbox.StringRequest;
-import com.android.volley.toolbox.Volley;
 import com.box.androidsdk.content.BoxApiFile;
 import com.box.androidsdk.content.BoxConfig;
 import com.box.androidsdk.content.BoxException;
@@ -109,8 +101,6 @@ import java.io.OutputStream;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
-import java.util.HashMap;
-import java.util.Map;
 import org.fossasia.phimpme.R;
 import org.fossasia.phimpme.base.PhimpmeProgressBarHandler;
 import org.fossasia.phimpme.base.RecyclerItemClickListner;
@@ -122,15 +112,20 @@ import org.fossasia.phimpme.gallery.activities.LFMainActivity;
 import org.fossasia.phimpme.gallery.util.AlertDialogsHelper;
 import org.fossasia.phimpme.gallery.util.ThemeHelper;
 import org.fossasia.phimpme.share.flickr.FlickrHelper;
+import org.fossasia.phimpme.share.imgur.ImgurPicUploadReq;
+import org.fossasia.phimpme.share.imgur.ImgurPicUploadResp;
 import org.fossasia.phimpme.share.tumblr.TumblrClient;
 import org.fossasia.phimpme.share.twitter.HelperMethods;
 import org.fossasia.phimpme.utilities.ActivitySwitchHelper;
 import org.fossasia.phimpme.utilities.Constants;
+import org.fossasia.phimpme.utilities.ImgurApi;
 import org.fossasia.phimpme.utilities.NotificationHandler;
+import org.fossasia.phimpme.utilities.RetrofitClient;
 import org.fossasia.phimpme.utilities.SnackBarHandler;
 import org.fossasia.phimpme.utilities.Utils;
-import org.json.JSONException;
-import org.json.JSONObject;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
 
 /**
  * Class which deals with Sharing images to multiple Account logged in by the user in the app. If
@@ -155,6 +150,9 @@ public class SharingActivity extends ThemedActivity
         OnRemoteOperationListener,
         RecyclerItemClickListner.OnItemClickListener {
 
+  private static final String IMGUR_BASE_URL = "https://api.imgur.com/3/";
+  private static final String IMGUR_HEADER_CLIENT = "Client-ID";
+  private static final String IMGUR_HEADER_USER = "Bearer";
   public static final String EXTRA_OUTPUT = "extra_output";
   private static String LOG_TAG = SharingActivity.class.getCanonicalName();
   public String saveFilePath;
@@ -202,9 +200,10 @@ public class SharingActivity extends ThemedActivity
   public String uploadName;
   private int positionShareOption;
   private boolean triedUploading = false;
+  private ImgurApi imgurApiInterface;
 
   public static String getClientAuth() {
-    return Constants.IMGUR_HEADER_CLIENt + " " + Constants.MY_IMGUR_CLIENT_ID;
+    return IMGUR_HEADER_CLIENT + " " + Constants.MY_IMGUR_CLIENT_ID;
   }
 
   @Override
@@ -1015,6 +1014,9 @@ public class SharingActivity extends ThemedActivity
   }
 
   private void shareToImgur() {
+    if (imgurApiInterface == null) {
+      imgurApiInterface = RetrofitClient.getRetrofitClient(IMGUR_BASE_URL).create(ImgurApi.class);
+    }
     final AlertDialog.Builder dialogBuilder =
         new AlertDialog.Builder(SharingActivity.this, getDialogStyle());
     RealmQuery<AccountDatabase> query = realm.where(AccountDatabase.class);
@@ -1022,7 +1024,7 @@ public class SharingActivity extends ThemedActivity
     final RealmResults<AccountDatabase> result = query.findAll();
     if (result.size() != 0) {
       isPersonal = true;
-      imgurAuth = Constants.IMGUR_HEADER_USER + " " + result.get(0).getToken();
+      imgurAuth = IMGUR_HEADER_USER + " " + result.get(0).getToken();
     }
     AlertDialogsHelper.getTextDialog(
         SharingActivity.this, dialogBuilder, R.string.choose, R.string.imgur_select_mode, null);
@@ -1081,21 +1083,30 @@ public class SharingActivity extends ThemedActivity
     Bitmap bitmap = getBitmapFromPath(saveFilePath);
     final String imageString = getStringImage(bitmap);
     // sending image to server
-    StringRequest request =
-        new StringRequest(
-            Request.Method.POST,
-            Constants.IMGUR_IMAGE_UPLOAD_URL,
-            new Response.Listener<String>() {
+    ImgurPicUploadReq imgurPicUpload = new ImgurPicUploadReq();
+    imgurPicUpload.setImage(imageString);
+    if (caption != null && !caption.isEmpty()) {
+      imgurPicUpload.setCaption(caption);
+    }
+    String authorization;
+    if (isPersonal && imgurAuth != null) {
+      authorization = imgurAuth;
+    } else {
+      authorization = getClientAuth();
+    }
+    imgurApiInterface
+        .uploadImageToImgur(authorization, imgurPicUpload)
+        .enqueue(
+            new Callback<ImgurPicUploadResp>() {
               @Override
-              public void onResponse(String s) {
-                dialog.dismiss();
-                JSONObject jsonObject = null;
-
-                try {
-                  jsonObject = new JSONObject(s);
-                  Boolean success = jsonObject.getBoolean("success");
+              public void onResponse(
+                  Call<ImgurPicUploadResp> call, Response<ImgurPicUploadResp> response) {
+                if (response.body() != null && response.isSuccessful()) {
+                  dialog.dismiss();
+                  ImgurPicUploadResp imgurPicUploadResp = response.body();
+                  boolean success = imgurPicUploadResp.isSuccess();
                   if (success) {
-                    final String url = jsonObject.getJSONObject("data").getString("link");
+                    final String url = imgurPicUploadResp.getData().getLink();
 
                     if (isPersonal) {
                       imgurString = getString(R.string.upload_personal) + "\n" + url;
@@ -1143,45 +1154,18 @@ public class SharingActivity extends ThemedActivity
                     SnackBarHandler.create(parent, getString(R.string.error_on_imgur)).show();
                     sendResult(FAIL);
                   }
-                } catch (JSONException e) {
-                  e.printStackTrace();
+                } else {
+                  dialog.dismiss();
+                  SnackBarHandler.create(parent, getString(R.string.error_volly)).show();
                 }
               }
-            },
-            new Response.ErrorListener() {
+
               @Override
-              public void onErrorResponse(VolleyError volleyError) {
+              public void onFailure(Call<ImgurPicUploadResp> call, Throwable t) {
                 dialog.dismiss();
-                SnackBarHandler.create(parent, getString(R.string.error_volly))
-                    .show(); // add volleyError to check error
+                SnackBarHandler.create(parent, getString(R.string.error_volly)).show();
               }
-            }) {
-          @Override
-          protected Map<String, String> getParams() throws AuthFailureError {
-            Map<String, String> parameters = new HashMap<String, String>();
-            parameters.put("image", imageString);
-            if (caption != null && !caption.isEmpty()) parameters.put("title", caption);
-            return parameters;
-          }
-
-          @Override
-          public Map<String, String> getHeaders() throws AuthFailureError {
-            Map<String, String> headers = new HashMap<String, String>();
-            if (isPersonal) {
-              if (imgurAuth != null) {
-                headers.put(getString(R.string.header_auth), imgurAuth);
-              }
-            } else {
-              headers.put(getString(R.string.header_auth), getClientAuth());
-            }
-
-            return headers;
-          }
-        };
-    request.setRetryPolicy(
-        new DefaultRetryPolicy(50000, 5, DefaultRetryPolicy.DEFAULT_BACKOFF_MULT));
-    RequestQueue rQueue = Volley.newRequestQueue(SharingActivity.this);
-    rQueue.add(request);
+            });
   }
 
   /**
--- a/app/src/main/java/org/fossasia/phimpme/share/imgur/ImgurPicUploadResp.java
+++ b/app/src/main/java/org/fossasia/phimpme/share/imgur/ImgurPicUploadResp.java
@@ -1,0 +1,52 @@
+package org.fossasia.phimpme.share.imgur;
+
+public class ImgurPicUploadResp {
+  private boolean success;
+  private Data data;
+
+  public boolean isSuccess() {
+    return success;
+  }
+
+  public void setSuccess(boolean success) {
+    this.success = success;
+  }
+
+  public Data getData() {
+    return data;
+  }
+
+  public void setData(Data data) {
+    this.data = data;
+  }
+
+  public static class Data {
+    private String id;
+    private String title;
+    private String link;
+
+    public String getId() {
+      return id;
+    }
+
+    public void setId(String id) {
+      this.id = id;
+    }
+
+    public String getTitle() {
+      return title;
+    }
+
+    public void setTitle(String title) {
+      this.title = title;
+    }
+
+    public String getLink() {
+      return link;
+    }
+
+    public void setLink(String link) {
+      this.link = link;
+    }
+  }
+}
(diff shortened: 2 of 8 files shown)


codedsun:
@iamareebjamal @yashk2000 Please merge this!

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -1081,107 +1079,95 @@ void uploadImgur() {
     Bitmap bitmap = getBitmapFromPath(saveFilePath);
     final String imageString = getStringImage(bitmap);
     // sending image to server
-    StringRequest request =
-        new StringRequest(
-            Request.Method.POST,
-            Constants.IMGUR_IMAGE_UPLOAD_URL,
-            new Response.Listener<String>() {
+    ApiInterface retrofitClient =
+        RetrofitClient.getRetrofitClient(Urls.IMGUR_BASE_URL).create(ApiInterface.class);
```
iamareebjamal: This should be done once and stored in a static variable

app/src/main/java/org/fossasia/phimpme/utilities/ApiInterface.java
```diff
@@ -0,0 +1,16 @@
...
+import com.google.gson.JsonElement;
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.Header;
+import retrofit2.http.POST;
+
+/** Created by @codedsun on 19/Oct/2019 */
+public interface ApiInterface {
+
+  @POST("/3/image")
```
iamareebjamal: /3/ should be part of the base URL

app/src/main/java/org/fossasia/phimpme/utilities/Urls.java
```diff
@@ -0,0 +1,7 @@
+package org.fossasia.phimpme.utilities;
+
+/** Created by @codedsun on 19/Oct/2019 */
+public class Urls {
+
+  public static final String IMGUR_BASE_URL = "https://api.imgur.com";
```
iamareebjamal: Not useful as it is not reused

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
codedsun: We can't because we need to have retrofit client with multiple base Url @iamareebjamal

iamareebjamal: > We can't because we need to have retrofit client with multiple base Url

Look at this, again
`RetrofitClient.getRetrofitClient(Urls.IMGUR_BASE_URL)`

Then look at this, again
`RetrofitClient.getRetrofitClient(Urls.IMGUR_BASE_URL).create(ApiInterface.class)`

And then read your comment

app/src/main/java/org/fossasia/phimpme/utilities/Urls.java
codedsun: It's used while creating retrofit client

iamareebjamal: https://en.wikipedia.org/wiki/Reuse

This is an interesting read on this topic

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
codedsun: @iamareebjamal I have created client for Imgur but when I want to create for any other api I have to pass that api base url, for example pinterest.

We can't create a singleton class and store in static variable when we have multiple base url.

app/src/main/java/org/fossasia/phimpme/utilities/ApiInterface.java
codedsun: Moved, done!

app/src/main/java/org/fossasia/phimpme/utilities/Urls.java
codedsun: @iamareebjamal We have one place for keeping all the API's URL.

codedsun: i will keep all the other accounts' url api endpoints here, if you want me to remove the file and use that directly in activity or something i will do that.

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
iamareebjamal: You can never use `ApiInterface` for any other base URL or even any other endpoint. If you can, I'll change my name to whatever you say

iamareebjamal: https://stackoverflow.com/questions/41672427/how-to-make-a-singleton-for-retrofit-2

I'll recommend you do some research before making wrong claims

codedsun: > Look at this, again
RetrofitClient.getRetrofitClient(Urls.IMGUR_BASE_URL)
Then look at this, again
RetrofitClient.getRetrofitClient(Urls.IMGUR_BASE_URL).create(ApiInterface.class)

I will have another retrofit client for uploading image to Pinterest, that's why i have followed this approach. PR is in pending state due to failing volley multipart request.

codedsun: @iamareebjamal I know that we can create retrofit instance in a singleton class but this can be done when we have baseUrl only 1 through out the app. But as the pineterest account uses another base url I will have to create another singleton class for that base url as you suggested..

what i followed to not follow singleton approach and build the retrofit client with the base URL provided in the function itself.

```ApiInterface``` is just an interface with different endpoints. You can invoke that function which corresponds to that baseUrl. It makes sense that with Imgur base URL i will not call uploadImagetoPinterest function and vice versa.


If i am wrong please guide me what shall I do.?

Comment (codedsun) [reply app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java]: You want me to have multiple retrofit client in the app and mutiple interfaces for endpoints, I will go with this approach..

**Comment by codedsun** [comment]
@iamareebjamal What i m getting is to have multiple services (API Endpoint) for every baseUrl and no Urls file as they are not re-used anywhere?

**Comment by codedsun** [comment]
@iamareebjamal You say to have a single retrofit instance for imgur, it's right but what shall I do when I have to hit with another base URL like for pinterest upload image, PR #2892. Then will I create another retrofit client. Then at runtime how will I change the base URL. Please define where I am wrong and what steps should i take to make the PR correct and merge ready.!

Thanks!

File: app/src/main/java/org/fossasia/phimpme/utilities/Urls.java (hunk 3 above)
Comment (iamareebjamal) [reply app/src/main/java/org/fossasia/phimpme/utilities/Urls.java]: That's a solid anti-pattern

File: app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java (hunk 1 above)
Comment (iamareebjamal) [reply app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java]: Retrofit instance creation uses reflection and is expensive and should be done only once per application or activity, not on each function call

**Comment by iamareebjamal** [comment]
You have to create another retrofit instance for another base URL anyway, so I don't get your point at all. Why are you creating the retrofit instance and the service class both on each call of the function?

Imagine retrofit instance and service class instantiation takes 3 seconds to complete for the sake of argument. And rest of the method takes 1 sec. Your method will take 4 second each time and my method will take 4 second for first time and 1 second for every other time.

File: app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java (hunk 1 above)
Comment (codedsun) [reply app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java]: @iamareebjamal Shall I create the retrofit instance in the activity itself and should I pass the URL from here or keep that where the retrofit instance is created?

Comment (iamareebjamal) [reply app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java]: Pass from here

**Comment by codedsun** [comment]
@iamareebjamal Convinced, thankyou for such detail. Please check now

File: app/src/main/java/org/fossasia/phimpme/utilities/ImgurRetrofitClient.java [hunk 4]
```diff
@@ -0,0 +1,40 @@
...
+    return httpClient;
+  }
+
+  public static Retrofit getImgurRetrofitClient() {
+    if (imgurRetrofitClient == null) {
+      imgurRetrofitClient = new ImgurRetrofitClient();
+    }
+    return imgurRetrofitClient.retrofit;
+  }
+}
```
iamareebjamal: No need to create an entire class and factory just for another Base URL

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -202,6 +199,8 @@
   public String uploadName;
   private int positionShareOption;
   private boolean triedUploading = false;
+  private ImgurApiInterface imgurApiInterface =
+      ImgurRetrofitClient.getImgurRetrofitClient().create(ImgurApiInterface.class);
```
iamareebjamal: Lazily intialize it

app/src/main/java/org/fossasia/phimpme/utilities/ImgurRetrofitClient.java
codedsun: means I should shift this code to activity?

codedsun: I have renamed the earlier class from retrofitClient to ImgurRetrofitClient, It follows singleton pattern, that's why a seperat class is created and also you suggested 

>Then you have to create another retrofit instance in any case! It has nothing to do with the fact that you are creating the service in method each time, instead of 1 time. You may use singleton or not, that is your choice. Both of those approaches will not save you from having to create another retrofit instance for pinterest upload

codedsun: @iamareebjamal Please guide me what should be done?

codedsun:
@iamareebjamal check now

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApiInterface.java
```diff
@@ -0,0 +1,16 @@
...
+
+import android.util.ArrayMap;
+import com.google.gson.JsonElement;
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.Header;
+import retrofit2.http.POST;
+
+/** Created by @codedsun on 19/Oct/2019 */
+public interface ImgurApiInterface {
```
iamareebjamal:
```suggestion
public interface ImgurApi {
```

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApiInterface.java
```diff
@@ -0,0 +1,16 @@
+package org.fossasia.phimpme.utilities;
+
+import android.util.ArrayMap;
+import com.google.gson.JsonElement;
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.Header;
+import retrofit2.http.POST;
+
+/** Created by @codedsun on 19/Oct/2019 */
```
iamareebjamal: Remove

app/src/main/java/org/fossasia/phimpme/utilities/RetrofitClient.java
```diff
@@ -0,0 +1,28 @@
...
+  private static OkHttpClient.Builder httpClient() {
+    OkHttpClient.Builder httpClient = new OkHttpClient.Builder();
+    HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
+    interceptor.level(HttpLoggingInterceptor.Level.BODY);
+    httpClient.interceptors().add(interceptor);
+    return httpClient;
+  }
+
+  public static Retrofit getRetrofitClient(String baseUrl) {
+    return new Retrofit.Builder()
```
iamareebjamal: Cache it

codedsun: @iamareebjamal for how much time?

iamareebjamal: Forever

codedsun: @iamareebjamal - Cache the response forever?

iamareebjamal: Sorry, I thought it was returning a static instance

codedsun: Will update the PR tomorrow and continue my work for Pintrest PR.

codedsun:
@iamareebjamal Check!

app/src/main/java/org/fossasia/phimpme/utilities/Constants.java
```diff
@@ -8,7 +8,6 @@
   public static final int REQUEST_SHARE_RESULT = 50;
   public static final String SHARE_RESULT = "share_result";
 
-  public static final String IMGUR_IMAGE_UPLOAD_URL = "https://api.imgur.com/3/image/";
   public static String IMGUR_HEADER_CLIENt = "Client-ID";
   public static String IMGUR_HEADER_USER = "Bearer";
```
iamareebjamal: If these are used in one place only, move them to that place as well

app/src/main/java/org/fossasia/phimpme/utilities/RetrofitClient.java
```diff
@@ -0,0 +1,28 @@
+package org.fossasia.phimpme.utilities;
+
+import okhttp3.OkHttpClient;
+import okhttp3.logging.HttpLoggingInterceptor;
+import retrofit2.Retrofit;
+import retrofit2.converter.gson.GsonConverterFactory;
+
+/** Created by @codedsun on 19/Oct/2019 */
```
iamareebjamal: Remove

app/src/main/java/org/fossasia/phimpme/utilities/RetrofitClient.java
```diff
@@ -0,0 +1,28 @@
...
+
+// Retrofit client used for API Calls
+public class RetrofitClient {
+
+  private static OkHttpClient.Builder httpClient() {
+    OkHttpClient.Builder httpClient = new OkHttpClient.Builder();
+    HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
+    interceptor.level(HttpLoggingInterceptor.Level.BODY);
+    httpClient.interceptors().add(interceptor);
+    return httpClient;
```
iamareebjamal: This is same for every instance, so it should be cached

app/src/main/java/org/fossasia/phimpme/utilities/RetrofitClient.java
```diff
@@ -0,0 +1,28 @@
...
+
+/** Created by @codedsun on 19/Oct/2019 */
+
+// Retrofit client used for API Calls
+public class RetrofitClient {
+
+  private static OkHttpClient.Builder httpClient() {
+    OkHttpClient.Builder httpClient = new OkHttpClient.Builder();
+    HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
+    interceptor.level(HttpLoggingInterceptor.Level.BODY);
```
iamareebjamal: Level BODY is too verbose

codedsun: @iamareebjamal Just used for development purpose, will add check of debug and production here from buildconfig

app/src/main/java/org/fossasia/phimpme/utilities/RetrofitClient.java
codedsun: These are automatically generated. Ok will remove

app/src/main/java/org/fossasia/phimpme/utilities/RetrofitClient.java
iamareebjamal: BODY is too verbose even for debug. Stetho should be used instead

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApi.java
```diff
@@ -0,0 +1,15 @@
...
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.Header;
+import retrofit2.http.POST;
+
+public interface ImgurApi {
+
+  @POST("/image")
+  Call<JsonElement> uploadImageToImgur(
+      @Header("Authorization") String authorization, @Body ArrayMap<String, String> body);
```
iamareebjamal: Change body to a class

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApi.java
```diff
@@ -0,0 +1,15 @@
...
+import com.google.gson.JsonElement;
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.Header;
+import retrofit2.http.POST;
+
+public interface ImgurApi {
+
+  @POST("/image")
+  Call<JsonElement> uploadImageToImgur(
```
iamareebjamal: JsonElement to as well

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApi.java
codedsun: Means?

codedsun: why not arraymap?

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApi.java
codedsun: ? Response to a class? why?

codedsun: Do we have a class for every API response? why should be follow this approach?

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApi.java
iamareebjamal: Because it's java, not python or JS

app/src/main/java/org/fossasia/phimpme/utilities/ImgurApi.java
iamareebjamal: See the extension of the file

codedsun: i will change but JsonElement is also a class present under  ```package com.google.gson;```

iamareebjamal: Obviously, it is. You can even use ArrayList without type parameters to make it like Python and JS. You can do anything in Java that you can do in Python or JS, but we want [type safety](https://en.wikipedia.org/wiki/Type_safety)

codedsun:
@iamareebjamal Review !

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -155,7 +150,10 @@
         OnRemoteOperationListener,
         RecyclerItemClickListner.OnItemClickListener {
 
+  private static final String IMGUR_BASE_URL = "https://api.imgur.com/3/";
   public static final String EXTRA_OUTPUT = "extra_output";
+  public static String IMGUR_HEADER_CLIENT = "Client-ID";
+  public String IMGUR_HEADER_USER = "Bearer";
```
iamareebjamal: Group IMGUR fields together

app/src/main/java/org/fossasia/phimpme/share/imgur/ImgurPicUploadResp.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public Data getData() {
+    return data;
+  }
+
+  public void setData(Data data) {
+    this.data = data;
+  }
+
+  public class Data {
```
iamareebjamal:
```suggestion
  public static class Data {
```

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -155,6 +150,9 @@
         OnRemoteOperationListener,
         RecyclerItemClickListner.OnItemClickListener {
 
+  private static final String IMGUR_BASE_URL = "https://api.imgur.com/3/";
+  public static String IMGUR_HEADER_CLIENT = "Client-ID";
+  public String IMGUR_HEADER_USER = "Bearer";
```
iamareebjamal: Why are these public and not static final?

iamareebjamal:
```suggestion
  private static final String IMGUR_HEADER_USER = "Bearer";
```

app/src/main/java/org/fossasia/phimpme/share/SharingActivity.java
```diff
@@ -155,6 +150,9 @@
         OnRemoteOperationListener,
         RecyclerItemClickListner.OnItemClickListener {
 
+  private static final String IMGUR_BASE_URL = "https://api.imgur.com/3/";
+  public static String IMGUR_HEADER_CLIENT = "Client-ID";
```
iamareebjamal:
```suggestion
  private static final String IMGUR_HEADER_CLIENT = "Client-ID";
```

codedsun:
@iamareebjamal are we good to go now?

codedsun:
@iamareebjamal Thank you for approval and efforts!

PR #2925, opened 2019-11-18
Fixed bottomBar Slide Animation

### LINKED ISSUE #2923: Bottom menu bar "move in effect" not from expected direction
If a user clicks on the left "camera" icon it would be expected that the cam function screen moves in from the left as well. If the user is on the "accounts screen" and clicks "Gallery" it would be expected that the "Gallery screen" moves in from the left. Instead it moves in from the right.

Please check and correct this. Screens should be moving in from the expected direction according to the bottom buttons.

satyajiit:
Fixed #2923 

Changes: Added a global variable to MyApplication Class.
And applied transition based on logic/conditions to each of the three activities.
Since we have three NavItems NavItem starts from zero to two.
Added four new anim xml files and deleted the old two.

GIF :
gif

Thank you @codedsun for your valuable feedback and review on this issue. 🙂


app/src/main/java/org/fossasia/phimpme/opencamera/Camera/CameraActivity.java
```diff
@@ -180,7 +181,7 @@ public void onCreate(Bundle savedInstanceState) {
     }
     super.onCreate(savedInstanceState);
     ButterKnife.bind(this);
-    overridePendingTransition(R.anim.right_to_left, R.anim.left_to_right);
+    startSlideAnimation(0);
```
satyajiit: Zero Marks the Start Position of NavItem i.e Camera

app/src/main/res/anim/left_to_right.xml
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
satyajiit: Not Required

rob729 (commented):
@satyajiit Squash these commits into a single commit

satyajiit:
@rob729 is it fine now?
Or should i start a new PR?

yashk2000 (approved):

codedsun (changes requested):
Use base activity instead of application class

